PR #6, opened 2023-09-19
Volunteer Sign Up Wireframe

### LINKED ISSUE #2: Volunteer sign-up wireframe in HTML
**What is needed**
The volunteer scheduling app will have various features, one of which a page allowing volunteers to sign up for shifts at shelters. Create a wireframe of such a sign-up page using HTML/CSS, including all the content you think this page should include. Use project requirements to identify the necessary content.

The wireframe should not use any colors other than grayscale. We want to focus on the content and the layout of that content.

**Additional context**
Read this article for more context on wireframes: https://www.codementor.io/@nicolesaidy/getting-started-with-wireframes-du107vuh7

loganwyas:
Closes #2

**What was changed?**
A wireframe for the volunteer sign-up widget was created using HTML.

**Why was it changed?**
It was changed because we want to get an idea of what this part of the site is going to look like.

**How was it changed?**
I created an HTML page, and wrote some HTML and CSS to create the wireframe.

image


kate-holdener:
I'm thinking the sign-up process is multi-step:
1. User clicks on some kind of "Sign up" button (probably from the volunteer dashboard)
2. User enters zip code (or "use current location")
3. A calendar is presented where the user selects date/time (the date/time options for when the user is already signed-up are disabled)
4. A list of locations is presented
5. User selects one of the listed locations
6. A confirmation page listing the selected date, time and location is presented
7. User confirms their choice and finalizes the sign-up.

loganwyas:
When I designed this, I assumed that there would be a calendar for each shelter the user wants to see (based on location, as mentioned). I wasn't sure if the shelters got to choose the specifications (like the length of each shift, how many people per shift, etc.), so I created it with the assumption that they had the choice. I could be wrong, though, and can change the design if necessary.

loganwyas:
@kate-holdener I have made changes based on our meeting today. The main things I added were text to show that sign-up is for an individual shelter, checkboxes on each of the available shifts to show they are selectable, text on the shifts to show its availability, a button to sign up for all of the selected shifts, and an alert to ask the user if they are okay with signing up for the selected shifts.

image

kate-holdener:
@loganwyas I wonder if it would be simpler if the shifts just showed up as a list instead of being a calendar. Some people may be doing this from their phone (so smaller screens). The list view would allow to more easily mark up which shift has a conflict (or maybe multiple shifts are in conflict with something that a volunteer already signed up for). The list view would also fit multi-week views on one page.

I also pictured this as a multi-step process: select shifts (page 1), review and confirm (page 2), Thanks for signing up for these shifts (page 3)

loganwyas:
That's true. I wasn't sure what exactly he wanted, but a list probably would work better. I can get started on that version of it, and we can look back on the branch later just in case he ends up wanting the calendar in the future.

kate-holdener:
@loganwyas You can probably leave this version as is, and just create another version that can co-exist along side with the old one. We can later remove the original version.

kate-holdener (approved):
I like the list version, I think it'll be easier to use.

kate-holdener:
Screenshot of the list version:

Screen Shot 2023-09-22 at 3 12 10 PM

PR #14, opened 2023-09-28
Common components

kate-holdener:
Defined some starting common components of the server side application, following the clean architectures in Python (second edition) book: https://leanpub.com/clean-architectures-in-python.

The current architecture follows the book up through and including chapter 4. For now, we only have an in-memory database. This will be changed later. The purpose of this pull request is to give the team common components and the starting architecture of the application.

The server-side exposes one endpoint /shifts, which returns a list of all work shifts present in the in-memory database (hard-coded to 2 shifts for now)

Client side was updated to display the received data:

Screen Shot 2023-09-28 at 11 34 09 AM


server/wsgi.py
```diff
@@ -0,0 +1,9 @@
+import os
+
+from flask_cors import CORS
+from application.app import create_app
+
+app = create_app(os.environ["FLASK_CONFIG"])
+
+CLIENT_URL = "http://localhost:3000" # URL of the React app
```
AbhilashKotha, line 8:
```suggestion
CLIENT_URLS = ["http://localhost:3000","http://127.0.0.1:3000"] URL of the React app
```

server/wsgi.py
```diff
@@ -0,0 +1,9 @@
...
+cors = CORS(app, resources={r"*": {"origins": CLIENT_URL}})
```
AbhilashKotha, line 9:
```suggestion
cors = CORS(app, resources={r"*": {"origins": CLIENT_URLS}})
```

AbhilashKotha (changes requested):
@kate-holdener I have suggested a change in CORS URLs. I have seen that causing issues in mac do you agree?

kate-holdener:
Thanks @AbhilashKotha! I added your suggested changes (and added real timestamps to our canned data)

AbhilashKotha (approved):
lgtm 👍  ; sorry for the typo.

PR #16, opened 2023-09-29
Add volunteer sign-up page

### LINKED ISSUE #7: Add a volunteer sign-up feature
Add a basic volunteer sign-up process, where the user has a way to specify a location+radius of where they'd like to volunteer, specify the time when they are available,  pick a shelter that's within the radius, and confirm the sign-up.

The basic sign-up process is a good place to start. In the future, we'll want to integrate more knowledge into this process. For example, we'll want to disable dates/times when the user is already known to have signed up for volunteer shifts.

The information about shelters should be coming from GetHelp API. The shifts that get created should store shelter's unique identifier, so that when we later retrieve these shifts later, we can retrieve the associated shelter information for GetHelp using shelter's unique id.

**Additional context**
Use volunteer sign-up wireframes in this repo (once available) to develop this feature

loganwyas:
Closes #7 

**What was changed?**
I added a page that lists all of the shelters and shifts associated with each shelter.

**Why was it changed?**
It was changed because we need a basic sign-up page before we can continue further.

**How was it changed?**
I added new files in the client to associate the pages with. I also modified the server to accept adding new WorkShifts.

image


kate-holdener:
@loganwyas Thanks for getting this started. Our starter-code in the front-end displays all the shifts (signed up shifts) in the database. I don't think that's where you need to be making the changes. 

The front end will need to display all time slots available for sign up (per shelter). You'll need to call GetHelp API for that. Then, the user will select time slots for sign up. That information will go to our server side, maybe something line /addWorkshift POST request. The request body will have the information about the time slots the volunteer wants to sign up for. This will then create WorkShift objects (one per time slot), with that volunteer attached to it. It might be easier to start on the server side.

server/application/rest/shelter.py
```diff
@@ -0,0 +1,29 @@
+import json
+
+import requests
```
AbhilashKotha, line 3: I think this needs to be added to requirements.txt

server/domains/shelter.py
```diff
@@ -0,0 +1,34 @@
...
+            "zipCode": d["zipCode"],
+            "latitude": d["latitude"],
+            "longitude": d["longitude"],
+            "beds": d["beds"],
+            "distance": d["distance"],
+        }
+        return self(**obj)
+
+    def to_dict(self):
+        return dataclasses.asdict(self)
```
kate-holdener, line 34: I don't think we want to be duplicating shelter info in the volunteer app. The call to get a list of shelters can come directly from the front end. The user will then pick a specific shelter and the request to create a WorkShift coming to our server would only have some kind of unique shelter id. We could validate that this id is real by making a call to GetHelp's API, then create the WorkShift using that ID as the shelter.

kate-holdener:
@loganwyas Will you be adding the server-side changes to create a WorkShift (based on the shelter/times that user selected)? Let me know if you need help getting started on the server-side.

loganwyas:
I plan to add it, I'm still working on the front-end logic at the moment.

server/application/rest/work_shift.py
```diff
@@ -31,17 +31,29 @@
...
+        result = workshift_list_use_case(repo)
+
+        return Response(
+            json.dumps(result, cls=WorkShiftJsonEncoder),
+            mimetype="application/json",
+            status=200,
+        )
+    elif request.method == "POST":
+        data = request.get_json()
+        shifts.extend(data)
```
kate-holdener, line 52: Instead of directly extending the data that we hard coded here for development purposes, you need to add a new use case for adding/creating a workshift. The usecase will be passed the repo object, and will call on the repo object to add the new workshift. That way, when we switch from using MemRepo to a real database, we will only need to instantiate a real database instead of MemRepo and the rest of the code will continue to work without changes.

client_app/src/Shelters.js
```diff
@@ -0,0 +1,114 @@
...
+  function manageShifts(shifts) {
+    setSelectedShifts(shifts);
+    setButtonDisabled(selectedShifts.length === 0);
+  }
+
+  function submitShifts() {
+    let shifts = selectedShifts;
+    for (let i = 0; i < shifts.length; i++) {
+      shifts[i].worker = "submitted-volunteer@slu.edu";
+    }
```
kate-holdener, line 65: We will know who the volunteer is because the volunteer will be signed in and will have a "bearer token". This "bearer token" is usually passed to the server in the header of an HTTPS packet. The server will then contact GetHelp's authentication server to figure out the email address associated with this bearer token. Since the authentication is not yet implemented, you should place (hard code) the volunteer email address into the packet header as the 'Authorization' field (you have headers defined below on line 70). Then, on the server side you can add a function 'getUserFromToken', which for now will just read the 'Authorization' field and return the email address. Later, when authentication is implemented, we will change getUserFromToken function to contact GetHelp's authentication server.

kate-holdener (changes requested):
@loganwyas I left a couple of comments, explaining what I think should be adjusted. Please review and let me know if you have any questions.

server/use_cases/add_workshifts.py
```diff
@@ -0,0 +1,14 @@
+"""
+This module contains the use case for adding work shifts.
+"""
+def workshift_add_use_case(repo, work_shift):
+    """
+    The function adds a workshift into the in-memory database.
```
kate-holdener, line 6: Please reword, to avoid references to in-memory database (this could be any data repository, we are just using in-memory repo for initial development)

server/use_cases/add_workshifts.py
```diff
@@ -0,0 +1,14 @@
...
+    """
+    repo.add(work_shift)
+def workshift_add_multiple_use_case(repo, work_shifts):
+    """
+    The function adds multiple workshifts into the in-memory database.
+    """
```
kate-holdener, line 12: Please reword, to avoid references to in-memory database (this could be any data repository, we are just using in-memory repo for initial development)

kate-holdener (changes requested):
@loganwyas Thanks for the changes. I made more comments requesting some more small, but important changes. We want to avoid duplication. This applies to both code and information we are passing around. That's the reasoning behind my latest comments.

kate-holdener (approved):
Looks good @loganwyas! Thanks for all the changes.

PR #26, opened 2023-10-10
Upcoming shifts component

### LINKED ISSUE #8: Create an 'upcoming shifts' component
The volunteer dashboard wireframe has a section showing all upcoming shifts I signed up for. That's a useful feature that needs to be implemented as a stand-alone component, which can then be included in the dashboard.

Our WorkShift data structure has a shelter identifier in it. This identifier should be used to lookup shelter information from the GetHelp API. Then, the Shelter data and WorkShift data can be combined and sent back as a response in JSON format. 

The request to the server should include user's authentication token in the packet header. This token can then be used to figure out the user information (email address) by contacting GetHelp authentication server. The request should return all shifts that are now or are in the future. Past shifts should not be included.

kate-holdener:
Fixes #8 

**What was changed**
The /shifts endpoint used to return all known work shifts in the database. Now it only returns shifts that belong to a specific user. Additionally, you can pass parameters such as filter_start_before, filter_start_after, filter_end_before, filter_end_after. The value of the parameter is a UTC time in milliseconds since the epoch. If specified, the value is used to filter the results and return only those work shifts that match the request. The 'after' filter is inclusive, the before filter is exclusive. 

**Why was it changed***
For security reasons, we want to limit the returned responses only to those shifts that were created by the user who is requesting the list. The parameters now give us a way to request upcoming/future shifts and past shifts. 

**How was it changed**
When processing the /shifts endpoint in application/rest/work_shift.py, we first identify if any parameter that starts with filter_ was passed. If so, we collect the parameters and their values and create a WorkShiftListValidRequest or WorkShiftListInvalidRequest object. The request object and the user information is passed to the workshift_list_use_case.

On the client side, Shifts.js now contains two exported functions: UpcomingShifts and PastShifts. These functions use current time to pass filter_start_after or filter_end_before parameters respectively to the /shifts endpoint.

The screenshot below shows the client side view, which loads upcoming shifts at the top and then shelters. The screenshot shows one shift for volunteer@slu.edu, which is in the future. We know that the server-side has two shifts for that user (one in the future and one in the past). 
Screen Shot 2023-10-10 at 6 08 26 PM

The data below shows 3 shifts, 2 of which are for volunteer@slu.edu. The first shift is in the past, the last shift is in the future (at the time of writing this comment).
```
  {
        "code": "f853578c-fc0f-4e65-81b8-566c5dffa35a",
        "worker": "volunteer@slu.edu",
        "shelter": "shelter-id-for-st-patric-center",
        "start_time": 1696168800000,
        "end_time": 1696179600000,
    },
    {
        "code": "f853578c-fc0f-4e65-81b8-566c5dffa35b",
        "worker": "volunteer2@slu.edu",
        "shelter": "shelter-id-for-st-patric-center",
        "start_time": 1696255200000,
        "end_time": 1696269600000,
    },
    {
        "code": "f853578c-fc0f-4e65-81b8-566c5dffa35a",
        "worker": "volunteer@slu.edu",
        "shelter": "shelter-id-for-st-patric-center",
        "start_time": 1701442800000,
        "end_time": 1701453600000,
    },
```


client_app/src/Shifts.js
```diff
@@ -29,5 +26,15 @@ function Shifts(){
...
+   const shelters_endpoint = SERVER+'/shifts?filter_start_after='+time_now;
+   return Shifts(shelters_endpoint)
+}
 
-export default Shifts;
+export function PastShifts(){
+   const time_now = new Date().getTime()
+   const shelters_endpoint = SERVER+'/shifts?filter_end_before='+time_now;
+   return Shifts(shelters_endpoint)
+}
```
loganwyas, line 40: Is this going to be implemented later into the volunteer dashboard? I saw it and didn't know where it was planned to be used in the code.

kate-holdener: I added it here to demonstrate how past shifts can be retrieved. This could be used in the dashboard later. I imagine volunteers may want to see when and where they have volunteered in the past. No immediate plans to include past shifts in the UI, but it's here if needed.

loganwyas (approved):
The code looks good to me. I'm assuming that it's best practice for me to merge it, correct? Just let me know if I need to.

kate-holdener:
I'd want to see what @AbhilashKotha thinks before we merge.

AbhilashKotha:
@kate-holdener can you include __init__.py files in requests and rest folders so that the import does not fail? I think that is OS specific but having it will not cause an issue

kate-holdener:
Thanks @AbhilashKotha. I added empty __init__.py files to all directories. Is that what you had in mind? I also updated how the response from the /shifts GET endpoint is built (to be consistent with the /shifts/<SHIFT ID> DELETE call)

AbhilashKotha (approved):
I think the code is ready to be merged. the status code and init files issue is addressed already.

PR #38, opened 2023-10-26
Issue 34 generate work shifts dynamically

### LINKED ISSUE #34: Shift date/time is hard-coded on the sign-up page
**Describe the bug**
When a volunteer wants to sign-up to help, the can see a list of shelters with dates/times of shifts that need help. However, those dates/times are currently hard-coded to:
October 03, 2023 12:00 - October 03, 2023 14:00
October 03, 2023 14:00 - October 03, 2023 16:00

This information should come from GetHelp API. The postman json data that was provided to us has information about PATCH request we can send to populate shelters' volunteer needs (although, more discussion with GetHelp may be needed. Not sure if their current API provides information about specific dates and times of work shifts that need help or not).

**To Reproduce**
1. Run the application (start mongodb, server, and client)
2. On the volunteer dashboard, look under 'shelters looking for volunteers').
3. You will see that all the dates and times are the same. They are hard-coded in client_app/src/Components/ShelterList.js

**Expected behavior**
Instead of hard-coded dates and times, the shift dates and times should be dynamically populated, from the information we receive from the GetHelp API.

**Screenshots**

Screen Shot 2023-10-22 at 3 47 07 PM

kate-holdener:
Fixes #34 

**What was changed?**

GetHelp API does not provide time slots for volunteer shifts per shelter. Therefore, the user will will be able to pick the start and end time of their shift. To assist the user in picking the start and ent time of a shift on a given day, the server generates information about number of workers signed up for different time intervals within that day. 

Added a new endpoint:
`/counts/<shelter-id>`

This endpoint requires two parameters:
```
filter_start_after
filter_end_before 
```

This endpoint will return a list of JSON objects specifying number of volunteers signed up for different time intervals throughout the given time period. The JSON object format is 
```
{
    "start_time" = int, 
    "end_time" = int, 
    "count" = int
}
```

**Why was it changed?**

To help volunteers make a decision on when to sign up 

**How was it changed?**

WIP: Server-side changes for now

Example request to this endpoint:
http://127.0.0.1:5000/counts/7343?filter_start_after=1696352400000&filter_end_before=1696366800000

Example response to this endpoint:
```
[
    {
        "start_time": 1696352400000,
        "end_time": 1696359600000,
        "count": 1
    },
    {
        "start_time": 1696359600000,
        "end_time": 1696366800000,
        "count": 2
    }
]
```


kate-holdener:
@loganwyas @AbhilashKotha I just requested your review on this. In particular, let me know if you can think of any other test cases I should add to tests/use_cases/test_count_volunteers.py

loganwyas (approved):
I think we have the right amount of test cases, as I can't think of any other kind of shift overlap

server/repository/mongorepo.py
```diff
@@ -31,37 +31,21 @@ def _create_shift_objects(self, results):
...
         Return a list of WorkShift objects based on the data.
         """
-        user_filter = {"worker":user}
+        db_filter = {}
+        if user:
+            db_filter["worker"] = user
+        if shelter:
+            db_filter["shelter"] = shelter
+
+        print(db_filter)
```
AbhilashKotha, line 44: I assume this was added during troubleshooting. Can we remove this?

server/tests/use_cases/test_count_volunteers.py
```diff
@@ -0,0 +1,358 @@
...
+            "end_time": 6
+        }
+    ]
+
+    request = Object(filters={"start_after":1, "end_before":4})
+    expected = [{"start_time":1, "end_time":4, "count":1}]
+
+    compare(expected, shifts_data, request)
+
+def test_workers_in_range_overlap():
```
AbhilashKotha, line 329: Could you please explain this to me? I am not understanding how duplicates are handled.

AbhilashKotha (commented):
I have tested the API manually and it works as expected and you have got good test cases. On top of that, I think test cases can be added for edge cases like when there are no shifts.

server/repository/mongorepo.py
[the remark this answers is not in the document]
kate-holdener: Good call, I removed this.

kate-holdener:
> I have tested the API manually and it works as expected and you have got good test cases. On top of that, I think test cases can be added for edge cases like when there are no shifts

Good idea, I'll add those test cases.

server/tests/use_cases/test_count_volunteers.py
kate-holdener: From our side discussion: there won't be duplicates in the database (same worker cannot be signed up for two overlapping shifts at the same time). I updated the test cases to make it clear that these are different workers that result in overlaps.

AbhilashKotha (approved):
lgtm;

Issue #48, opened 2023-11-07
Disallow selecting times from the past


kate-holdener:
**Describe the bug**
Our current client-side application allows users to select the start and end time of their work shift, including times from the past (see screenshot below, where the selected time is from year 2022, while it's 2023 right now). The user interface should restrict time selection (start and end time) to only future times.
<img width="1404" alt="Screen Shot 2023-11-07 at 12 10 30 PM" src="https://github.com/oss-slu/shelter_volunteers/assets/42073871/7a867e11-6ac8-40fd-a3f6-6bf6ef26d926">

**To Reproduce**
Run the database, server side application, and the client side. Use the client side "Sign up for shifts" feature to sign up for a shift from the past time.

**Expected behavior**
Only future dates/times should be allowed as starting and ending time of a shift. Going back in time should be disabled.

kate-holdener:
All yours @lirieh. Related to this, it would also be good to ensure that 'end time' cannot be before 'start time'. It would probably require similar code, so a judgement call if you want to fix it in this issue, or if we should track it separately.

lirieh:
The existing code already had a fix to ensure that the 'end time' is always after the 'start time'. However, this is done by always changing the start time. For example, if the current start time is 3:00 pm and you try to change the end time to 2:00 pm, the system changes the start time to 2:00 pm and the end time becomes 3:00 pm. I think a better solution in this case could be to change the end time to 2:00 pm (as per the user request) and the start time to 1:00 pm. If it is ok with you I can change the logic to work like this before submitting my code.

kate-holdener:
@lirieh let’s just not worry about the end time as compared to the start time for now (we can deal with it in a separate issue, as I think it deserves more discussion).

PR #49, opened 2023-11-12
Fix issue #48-Disallow selecting past times

lirieh:
The issue I selected was the current system allowed users to schedule times in the past (issue no 48). Therefore the issue to be fixed was to disallow users from selecting times in the past. 

**What was changed?**
All changes were made in the IndividualShelter.js file found in _client\_app/src/Components/_.

**Why was it changed?**
The calendar allowed users to select dates and times that were in the past. Therefore, changes were made to ensure that users could not select dates/times from the past in the calendar. The main changes are summarized below.

1. **_Show Available Dates_** - Updated the calendar to only shows dates from the current month onwards and only dates from the current date onwards can be selected.
2. **_Show Available Times_** - Updated the calendar to only show available start and end times. Available start times begin from the next half hour, and available end times begin 1 hour after the available start time.

**How was it changed?**
The main changes made in  _IndividualShelter.js_ file are summarized below:

1. _**Show Available Dates**_ - In the file, changes were made to the return component Start Time and End Time DatePickers. Added minDate (line 81, 93) to set the current date as the minimum active date to be shown in the calendar. Added showDisabledMonthNavigtion (line 82, 94) to set the current month as the oldest month in the calendar but with days older than today disabled.

3.  **_Show Available Times_** - Added filterPastStartTime (line 16) and filterPastEndTime (line 22) to show only available times for the Start Time and End Time calendars. The filterPastStartTime shows available times starting in the next half hour while filterPastEndTime shows available end times starting 1 hour after the first available start time. Added lines 77 and 89 in the return component to run these components.4. 
Fix issue 48

### DIFF
--- a/client_app/src/Components/IndividualShelter.js
+++ b/client_app/src/Components/IndividualShelter.js
@@ -13,6 +13,19 @@ const IndividualShelter = (props) => {
     setHours(setMinutes(new Date(), 0), new Date().getHours() + 2)
   );
 
+    const filterPastStartTime = (time) => {
+      const currentDate = new Date();
+      const selectedDate = new Date(time);
+      return  currentDate.getTime() < selectedDate.getTime();
+      };
+        
+    const filterPastEndTime = (time) => {
+       const currentDate = new Date();
+       const selectedDate = new Date(time);
+       currentDate.setHours(currentDate.getHours() + 1);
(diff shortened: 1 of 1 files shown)


kate-holdener (approved):
@lirieh These changes work well! Thank you.

PR #53, opened 2023-11-27
Make URL for GetHelp API configurable

### LINKED ISSUE #42: Make the URL of the 3rd party API (GetHelp) configurable
To retrieve shelter information, we call GetHelp's API. Currently, the URL of that API is hard coded to https://api2-qa.gethelp.com/. When the application is released to production, we will need to change the URL we are using to call the API.  This problem exists on both the client and server sides.

Rather than having to change it throughout the code, the URL should be configurable. This will allow us to simply change configuration.

loganwyas:
Fixes #42

**What was changed?**

I configured both the client and the server so that the URL for the GetHelp API is not completely hardcoded

**Why was it changed?**

This was changed because the URL will eventually need to be different when deploying to production

**How was it changed?**

On the client side, I added a variable to the config file that stores the URL and can be imported into whichever page wants to use it. On the server side, I added an environment variable to the testing and development config that is eventually used in some of the use cases.

### DIFF
--- a/client_app/src/Shelters.js
+++ b/client_app/src/Shelters.js
@@ -1,14 +1,16 @@
 import { useState, useEffect } from "react";
 import ShelterList from "./Components/ShelterList";
 import ConfirmationPage from "./Components/ConfirmationPage";
-import { SERVER } from "./config";
+import { GETHELP_API, SERVER } from "./config";
 import { Link } from "react-router-dom";
 import ShiftList from "./Components/ShiftList";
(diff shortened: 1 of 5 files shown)


kate-holdener (changes requested):
In server/use_cases/authenticate.py, there's a direct reference to 'https://api2-qa.gethelp.com/v1/users/current'

kate-holdener (approved):
Looks good, @loganwyas

PR #61, opened 2023-12-07
Map view for shelters

### LINKED ISSUE #24: Create a map view

Create a map view for the shelters nearby.

AbhilashKotha:
Fixes #24 
**What was changed?**

Users can now see the shelters on a map

**Why was it changed?**

To make it easy to the users to see the nearby shelters.

**How was it changed?**

Using Google Maps API

image


kate-holdener (changes requested):
I think it would be good to only have the "show map" option if there is a google maps api key configured. I'm thinking this would be more friendly for other developers contributing to this project. We wouldn't want to share our API key with everyone, but having the "show map" that results in an error (because there is no key) would result in extraneous bug reports.

kungfuchicken:
@kate-holdener is this still relevant? @AbhilashKotha are you still working this?

PR #68, opened 2024-02-02
Unique key prop warning for shelterList issue(#65) fixes 

### LINKED ISSUE #65: client_app - each child in a list should have a unique key prop
**Describe the bug**
When I run the client_app and navigate to the dashboard, I see a warning in the console stating "each child in a list should have a unique key prop".

**To Reproduce**
1. Run the application and sign in with [developer@slu.edu](mailto:developer@slu.edu) and any password
2. In your web browser, open "Developer tools"
3. In Developer tools, open the "console" tab
4. You should see "each child in a list should have a unique key prop" warning logged to the console. See screenshot below

Screenshot 2024-01-30 at 11 08 59 AM

**Expected behavior**
It's not exactly clear if this warning is hurting anything, but I would expect to not have console warnings.

**Additional context**
Investigate the problem and make the necessary code changes to remove this warning, without impacting the existing functionality.

cbiddle3:
Fixes #65

**What was changed?**

I changed the list of data (ShelterList) that is displayed on the homepage of the website. 

**Why was it changed?**

There was an issue about how each child should have a unique key. JSX auto-defaults to using index, but that can cause unexpected behavior when a list is not static and a longer rendering time. This issue appeared on the console when on the home page, as it was from the file that displayed the list of shelters (ShelterList.js)

**How was it changed?**

I added that when the shelter would get added if it met the condition in the if loop it would assign a key to the element using the unique shelter.id variable. 

Screenshot 2024-02-02 at 12 31 01 AM

Added line 25 and now the error is gone.

### DIFF
--- a/client_app/src/Components/ShelterList.js
+++ b/client_app/src/Components/ShelterList.js
@@ -22,7 +22,7 @@ const ShelterList = (props) => {
                 props.loadingFunction(false);
               }
               return (
-                <div>
+                <div key={shelter.id}>
                   <IndividualShelter
                     shelter={shelter}
                     isSignupPage={props.isSignupPage}


SiriChandanaGarimella:
@cbiddle3 - I see that you are on the right track. The changes you made to ShelterList.js, fix it on the Shelters page, but some other places also need the same fix. For example, shift list (upcoming and previous shifts page), table views (add shifts and shift confirmation page), etc. It would be great if you could do the other changes as part of this PR, else we can raise a new issue if you feel that it is too much work for this one.

The dashboard after adding a shift shows the error(due to the issue in the shift list)

image

SiriChandanaGarimella (approved):
lgtm;

Issue #64, opened 2024-01-30
client_app - cannot update a component


kate-holdener:
**Describe the bug**
When I run the client_app and navigate to the dashboard, I see a warning in the console stating "Cannot update a component".

**To Reproduce**
1. Run the application and sign in with [developer@slu.edu](mailto:developer@slu.edu) and any password
2. In your web browser, open "Developer tools"
3. In Developer tools, open the "console" tab
4. You should see a "Cannot update a component" warning logged to the console. See screenshot below
<img width="1313" alt="Screenshot 2024-01-30 at 11 01 23 AM" src="https://github.com/oss-slu/shelter_volunteers/assets/42073871/0174d770-761d-4a76-9c93-50f279e9e789">

**Expected behavior**
It's not exactly clear if this warning is hurting anything, but I would expect to not have console warnings.

**Additional context**
Investigate the problem and make the necessary code changes to remove this warning, without impacting the existing functionality.

PR #69, opened 2024-02-02
fixed issue 64 - Cannot Update a Component

tnhi26990:
Fixes issue #64 

**What was changed?**

I altered the shelter list on the website's home page. 

**Why was it changed?**

There was a warning in the console stating that the Shelters component cannot be updated while it is still in the process of rendering a different component (ShelterList). The goal is to make this warning disappear when the client_app is run. 

**How was it changed?**

The if statement in the return block is moved into a useEffect hook to ensure that the state update runs after the initial render. 

Screenshot 2024-02-02 at 5 31 50 PM
Screenshot 2024-02-02 at 5 32 19 PM

### DIFF
--- a/client_app/src/Components/ShelterList.js
+++ b/client_app/src/Components/ShelterList.js
@@ -1,11 +1,20 @@
 import IndividualShelter from "./IndividualShelter";
+import { useEffect } from "react";
 
 const ShelterList = (props) => {
-  function addShift(shift) {
-    if (props.manageShiftsFunction) {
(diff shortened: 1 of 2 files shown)


SiriChandanaGarimella:
@tnhi26990 - Great work! Just a small change, could you please update the title of the PR to be more detailed about what it is fixing,
Example: "Cannot Update a Component warning fixed"

SiriChandanaGarimella (approved):
lgtm;

PR #80, opened 2024-02-19
Fixes the unique shift code issue (#22), warning for unique key prop for list (#70) and warning for text-align in login page

### LINKED ISSUE #22: WorkShift id should be server-generated
**Describe the bug**
With the resolution of issue #7, volunteers have a way to sign up for shifts. However, the WorkShift ID is currently coming from the client side application: it is the shelter number followed by 'ffff'. It's coming from here: https://github.com/oss-slu/shelter_volunteers/blob/36bec9e8bd5c2a2fda4696399d8bcd651c303d16/client_app/src/Components/ShelterList.js#L57C21-L57C21

This means that our WorkShift IDs are not unique. These WorkShift IDs should be generated by the server side application and each created WorkShift should be assigned a unique value.

**To Reproduce**
1. Go to the server directory and start the server using README instructions
2. Go to the client_app directory and start the client using REAMDE instructions
3. Sign-up for different shifts within the same sheltert
4. Open a web browser window and point it to http://localhost:5000/shifts
5. You will see that the code associated with both new WorkShifts is the same

**Expected behavior**
WorkShift code should be unique. Typically, the database will generate a unique code for each newly added record.

### LINKED ISSUE #70: client_app - each child in a list should have a unique key prop (for all the lists)
**Describe the bug**
When I run the client_app and navigate to the Dashboard(with shifts added), Current selection page, Upcoming shifts, Previous shifts, or Shift confirmation page, I see a warning in the console stating "Each child in a list should have a unique key prop".

- In continuation to the issue #65
- Depends on the issue #22 

**To Reproduce**

1. Run the application and sign in with [developer@slu.edu](mailto:developer@slu.edu) and any password
2. In your web browser, open "Developer tools"
3. In Developer tools, open the "console" tab
4. Navigate to a page where a list or table is displayed (Dashboard(with shifts added), Current selection page, Upcoming shifts, Previous shifts, or Shift confirmation page).
4. You should see "Each child in a list should have a unique key prop" warning logged to the console. See screenshot below
image

**Expected behavior**
The console warning should not be displayed

**Additional context**
A unique key has to be added to each element in the list.

SiriChandanaGarimella:
Fixes #22 #70. Also, there is a warning for text-align on the login page

**What was changed?**
1. Generated a unique UUID and assigned the value to shift code in the IndividualShelter page. This is to resolve #22 
2. Added the key as shift.code for the table data on the ShiftList page and in the ConfirmationPage. This is resolve #70
3. Updated the style from text-align to textAlign in the About page. This resolves the warning Unsupported style property text-align on the sign-in page.

**Why was it changed?**
1. This is changed to make the shift code a unique value. As this is being used on different pages to uniquely identify each child in a list with a key prop, I made the changes in the client app instead of the server. 
2. As shift.code is a unique value, added it as a key prop for shifts table data.
3. As the accepted style is textAlign, I updated it from text-align to textAlign.

**How was it changed?**
1. Using the v4 method of the uuid package, I generated a unique UUID and appended it with the shelter ID, and assigned it to shift {code}
2. Added the key prop as shift.code for the div tag in ShiftList page and for the tr in ConfirmationPage.

**Issue 2 Fixes:**
image
image
image
image
image

**Issue 3 Fixes:**
image


client_app/src/Components/IndividualShelter.js
```diff
@@ -38,14 +38,15 @@ const IndividualShelter = (props) => {
...
+  const { v4: uuidv4 } = require('uuid');
 
   function addShift() {
     if (props.addShiftFunction) {
       let id = shelter.id;
       let start = startTime.getTime();
       let end = endTime.getTime();
       let shift = {
-        code: `${id}-${start}-${end}`,
+        code: `${uuidv4()}-${id}`,
```
kate-holdener, line 49: It seems to me that the unique code should be generated by the server/database, rather than the client side.

kate-holdener (changes requested):
@SiriChandanaGarimella I left a review comment for you. Also, you'll want to sync up this branch with the main branch (since the main branch moved on). My fault for letting this sit so long without reviewing it.

SiriChandanaGarimella:
@kate-holdener - I updated the code to use the unique ID generated by the database as the shift ID. Also, I made some changes in the front end and unit test cases accordingly. Could you please review it? Thanks!

server/repository/mongorepo.py
kate-holdener, line 1: @SiriChandanaGarimella I tested your code and it works well. It would be helpful if you could comment within mongodb.py how/where the unique id gets generated. It would be helpful for future developers.

SiriChandanaGarimella:
@kate-holdener - Thanks for the feedback! I've added comments to mongorepo.py to clarify where the unique ID gets generated. Please let me know if there are any modifications required.

server/repository/mongorepo.py
```diff
@@ -41,27 +42,25 @@ def list(self, user=None, shelter=None):
...
         return user_shifts
 
     def add(self, work_shift):
         """
         Add a WorkShift object to the data.
+        A unique id gets generated in mongoDB and is added to workShift object
         """
-        self.collection.insert_one(work_shift)
         work_shift.pop("_id")
+        self.collection.insert_one(work_shift)
```
kate-holdener, line 55:
```suggestion
        # Remove the _id field from the dictionary, so that when we call insert_one
        # mongoDB will add an _id field to the work_shift dictionary object with a unique
        # value before inserting it into the collection. This ensures that each work shift
        # stored in our database has a unique _id
        work_shift.pop("_id")
        self.collection.insert_one(work_shift)
```

kate-holdener (approved):

Issue #74, opened 2024-02-15
Make the API call to display the details in "Current Volunteer Counts" div on button click


SiriChandanaGarimella:
**Is this task related to a feature or a bug? Please include relevant issue numbers below.**
This task contributes to resolving issue #71 

**Describe the solution you'd like**
(Read issue description of #73 for more context)
Make the API call to display the details in the "Current Volunteer Counts" div only when the "View Current Volunteer Counts" button is clicked or "Start time date picker" or "End time date picker"
1. Remove the functionality to fetch the details in the "Current Volunteer Counts" div for each shelter through API call.
2. Make an API call to get the shift details of the shelter only when the "View Current Volunteer Counts" button, "Start time date picker" or "End time date picker" is clicked.
3. Also, make an API call when the value is updated in the "Start time date picker"

PR #83, opened 2024-03-02
Make the API call to display the details in "Current Volunteer Counts" div on button click (#74) fixed

tnhi26990:
Fixes #issue_74

**What was changed?**

I changed the the Volunteering Opportunities page. When a user clicks on the "View Current Volunteer Counts" button, the "Start time date" picker or the "End time date" picker in each shelter, an API call is made to get the shift details of that shelter only. 
 
**Why was it changed?**

It was changed to reduce a huge number of API call that could slow down the network. 

**How was it changed?**

I changed the condition of the if statement inside the useEffect hook to make API calls only when the state of the volunteerCountsHidden and shelter is changed.

ViewCVCisclicked
                                                      The "View Current Volunteer Counts" button is clicked.
STPisclicked
                                                                      The "Start time date" picker is clicked.
ETPisclicked
                                                                       The "End time date" picker is clicked.
STPischanged
                                                                          The "Start time picker" is updated.

### DIFF
--- a/client_app/src/Components/IndividualShelter.js
+++ b/client_app/src/Components/IndividualShelter.js
@@ -21,7 +21,7 @@ const IndividualShelter = (props) => {
   );
   const [shiftCounts, setShiftCounts] = useState([]);
   const [loading, setLoading] = useState(false);
-  const [volunteerCountsHidden, setVolunteerCountsHidden] = useState(true);
+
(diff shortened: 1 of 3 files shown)


SiriChandanaGarimella (approved):
LGTM

PR #89, opened 2024-03-22
Bug: inconsistent timestamps fixed (#62)

### LINKED ISSUE #62: Client is sending incorrect timestamps for new shifts
**Describe the bug**
When signing up for a shift, I think the client side is sending incorrect start/end times. Not sure exactly which time is incorrect, but there is definitely an inconsistency. I used console.log to print the timestamps being sent, and here's what I see.

Shift1:  12/10/2023 4pm - 5pm, the timestamps are:
start: 1702245634971
**end:  1702249234971**

Shift2:  12/10/2024 5pm - 6pm, the timestamps are:
**start: 1702249200000**
end:  1702252800000

This is an inconsistency, I would expect that the end time of shift 1 would be the same as the start time of shift 2.

**To Reproduce**
1. Add console.log statement to src/components/IndividualShelter in function addShift to print the start and end times
2. Start the server
3. Start the client
4. Sign up for shift 1
5. Sign up for shift 2, where the start of shift 2 equals to the end of shift 1
6. Examine the console output

**Expected behavior**
The start and end times of each shift should be the milliseconds since the Epoc in UTC (although, the times shown on the client should be in client local time).

cbiddle3:
Fixes #62 

**What was changed?**

Modified individualShelter.js

**Why was it changed?**

The timestamps being recorded for consecutive shifts weren't consistent (the time stamp given for the end of a shift at 3 pm and the start of another shift at 3 pm were different). 

**How was it changed?**

On individualShelter.js I changed lines 17 and 20 to set the milliseconds and seconds to 0. The new Date() wasn't getting the milliseconds and seconds attribute set to 0, so it would have the current difference in time from the milliseconds and seconds accounted into the timestamp. However natively for choosing times ahead (not the ones initially displayed on the page) minutes, seconds, and milliseconds are set to 0. Now the first displayed page time reflects that.

### DIFF
--- a/client_app/src/Components/IndividualShelter.js
+++ b/client_app/src/Components/IndividualShelter.js
@@ -14,10 +14,10 @@ import { faChevronDown, faChevronUp } from "@fortawesome/free-solid-svg-icons";
 const IndividualShelter = (props) => {
   let shelter = props.shelter;
   const [startTime, setStartDate] = useState(
-    setHours(setMinutes(new Date(), 0), new Date().getHours() + 1)
+    setHours(setMinutes(setSeconds(setMilliseconds(new Date (), 0), 0), 0), new Date().getHours() + 1)
   );
   const [endTime, setEndDate] = useState(
-    setHours(setMinutes(new Date(), 0), new Date().getHours() + 2)
+    setHours(setMinutes(setSeconds(setMilliseconds(new Date (), 0), 0), 0), new Date().getHours() + 2)
   );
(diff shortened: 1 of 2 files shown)


SiriChandanaGarimella:
@cbiddle3 - The code works fine. If possible, could you please add the unit test cases to test this functionality in the server? If not, I will create another issue for the test cases.

SiriChandanaGarimella (approved):
lgtm;

PR #91, opened 2024-03-23
Need clear error reporting when signing up for overlapping shifts (#88) fixed

### LINKED ISSUE #88: Need clear error reporting when signing up for overlapping shifts
**Describe the bug**
Using the client side of the application, I can select some shifts that overlap in time (user error) and proceed with the sign up process. The server correctly identifies the problem and only signs me up for one of the shifts, rejecting overlapping shifts and returning an error code 409. However, from the client side, it looks as if the sign-up was successful. See screenshots below.

I select overlapping times:
Screenshot 2024-03-20 at 4 00 30 PM

I submit the request and get this screen, making me think that I signed up for 2 shifts at different shelters:
Screenshot 2024-03-20 at 4 00 59 PM

When I look at my upcoming shifts, I see only one shift (which is correct, because the server only signed me up for one shift):
Screenshot 2024-03-20 at 4 53 25 PM

**To Reproduce**
1. Go to "Sign Up for Shifts"
2. Sign up for two shifts in different shelters with the same start and end times
3. Click on Submit Shifts, you'll see the screen that makes you think that you signed up for both shifts
4. Go to "Upcoming Shifts" and you'll see that you only signed up for one shift

**Expected behavior**
On step 3 above, the confirmation screen should somehow indicate which shifts were successfully created and which were not.

tnhi26990:
Fixes #88

**What was changed?**

I changed the Confirmation page by adding the fourth column titled "Status" to inform the user about the success status of each shift they signed up for. 

**Why was it changed?**

When the user signs up for overlapping shifts, there is no clear indication telling them that overlapping shifts are rejected.

**How was it changed?**

I changed Shelter.js and ConfirmationPage.js. In Shelter.js, I created a successList variable, and inside the submitShifts function, setSuccessList is used to update the successList state with the success value received from the Json response. I then added the successList as another condition to render the ConfirmationPage component and also passed the successList props to it. In ConfirmationPage.js, I added the fourth column labeled Status to the table header. I used successList[index] to determine whether the shift at the current index was successful or not. 

Screenshot 2024-04-03 at 3 28 04 PM
Screenshot 2024-04-03 at 3 28 49 PM

### DIFF
--- a/client_app/package-lock.json
+++ b/client_app/package-lock.json
@@ -8,9 +8,13 @@
       "name": "client_app",
       "version": "0.1.0",
       "dependencies": {
+        "@emotion/react": "^11.11.4",
+        "@emotion/styled": "^11.11.0",
         "@fortawesome/fontawesome-svg-core": "^6.4.2",
         "@fortawesome/free-solid-svg-icons": "^6.4.2",
         "@fortawesome/react-fontawesome": "^0.2.0",
+        "@mui/icons-material": "^5.15.14",
+        "@mui/material": "^5.15.14",
         "@react-spring/web": "^9.7.3",
         "@testing-library/jest-dom": "^5.17.0",
         "@testing-library/react": "^13.4.0",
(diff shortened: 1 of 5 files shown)


SiriChandanaGarimella (approved):
lgtm;

PR #128, opened 2024-09-28
Shelter Dashboard: Implement the 'Contact Past Volunteers' feature (#124) fixed

### LINKED ISSUE #124: Shelter Dashboard: Implement the 'Contact Past Volunteers' feature
Currently, the data in 'Contact Past Volunteers' is static. Fetch the data from the database and update the details in Contact Past Volunteers component.

Use the existing API to get the information on the volunteer list and fetch the data as required.
API: http://127.0.0.1:5000/getvolunteers/30207filter_start_after=1714107600000&filter_end_before=1714193999999
Currently, when the Shelter Dashboard is loaded, the above API is called on each interval in a day, so you only need to pass the list to the Contact Past Volunteers component to get the names displayed. 

**Acceptance Criteria:**

- [x] The "Contact Past Volunteers" component should display real-time volunteer data fetched from the database using the provided API.
- [x]  The API should only be called once during dashboard loading, and the data should be passed to the component as a prop or through context.
- [x]  Proper error handling should be implemented if the API call fails, displaying an appropriate message in the UI.
- [x]  All displayed volunteer names should be properly formatted and sorted based on the fetched data.
- [x]  The solution should be tested for responsiveness and ensure the list displays correctly on different screen sizes.

Screenshot 2024-09-23 at 9 00 57 PM

tnhi26990:
Fixes #124 

**What was changed?**

I needed to change the Contact Past Volunteers component in the Shelter Dashboard.

**Why was it changed?**

Currently, the data in the Contact Past Volunteers is static, and the data needs to be fetched from the database.

**How was it changed?**

I moved all the code calling the API on each time interval in the ShiftContainer.js to ShelterDashboard.js since we needed to pass the list to different components on the Shelter Dashboard page. The ShiftDetails list was passed to the ShiftContainer and PastVolunteersContainer components. Then the list was passed to the Roster and PastVolunteers component. I added code to Roster.js to show an additional message when the component hadn't finished loading. I added code to PastVolunteers to filter the list by start_time < current_time, take only the worker field, flat the array and remove all the punctuation marks. Then I used Set to remove duplicate names and then take the top three names from the array. I also added code to show an additional message when the component hadn't finished loading.

I removed unwanted code in the Shelter.js. I noticed the code using to print out the Shift Status List wasn't removed.

Screenshot 2024-09-28 at 12 49 35 PM
Screenshot 2024-09-28 at 12 50 36 PM

### DIFF
--- a/client_app/src/components/shelter/PastVolunteers.js
+++ b/client_app/src/components/shelter/PastVolunteers.js
@@ -3,17 +3,15 @@ import NotificationsNoneIcon from "@mui/icons-material/NotificationsNone";
 import "../../styles/shelter/PastVolunteers.css";
 
 const PastVolunteers = ({shiftDetails}) => {
-  
     const currentTime = Date.now();
-    const filteredShifts = shiftDetails.filter(item => item.start_time >= currentTime);
+    const filteredShifts = shiftDetails.filter(item => item.start_time < currentTime);
     const workerList = filteredShifts
-    .slice(0, 3)
     .flatMap(item => item.worker ? item.worker.split(",").map(name => name.trim()) : []);
     const uniqueWorker = [...new Set(workerList)];
 
     return (
       <div className="past-volunteers-container">
-        {uniqueWorker.length > 0 ? (uniqueWorker.map((name, index) => (
+        {uniqueWorker.length > 0 ? (uniqueWorker.slice(0,3).map((name, index) => (
           <div key={index} className="volunteers-item-container"> 
             <div className="volunteers-item">
               <AccountCircleIcon />
--- a/client_app/src/components/shelter/PastVolunteers.js
+++ b/client_app/src/components/shelter/PastVolunteers.js
@@ -1,22 +1,32 @@
-import React, { Component } from "react";
 import AccountCircleIcon from "@mui/icons-material/AccountCircle";
 import NotificationsNoneIcon from "@mui/icons-material/NotificationsNone";
 import "../../styles/shelter/PastVolunteers.css";
 
-class PastVolunteers extends Component {
-  render() {
-    const { shiftDetails } = this.props;
+const PastVolunteers = ({shiftDetails}) => {
+  
+    const currentTime = Date.now();
+    const filteredShifts = shiftDetails.filter(item => item.start_time >= currentTime);
+    const workerList = filteredShifts
+    .slice(0, 3)
+    .flatMap(item => item.worker ? item.worker.split(",").map(name => name.trim()) : []);
+    const uniqueWorker = [...new Set(workerList)];
 
     return (
-      <div className="volunteers-item-container">
-        <div className="volunteers-item">
-          <AccountCircleIcon />
-          <span>{shiftDetails.label}</span>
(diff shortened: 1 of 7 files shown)


client_app/src/components/shelter/PastVolunteers.js
```diff
@@ -1,22 +1,30 @@
...
 import AccountCircleIcon from "@mui/icons-material/AccountCircle";
 import NotificationsNoneIcon from "@mui/icons-material/NotificationsNone";
 import "../../styles/shelter/PastVolunteers.css";
 
-class PastVolunteers extends Component {
-  render() {
-    const { shiftDetails } = this.props;
+const PastVolunteers = ({shiftDetails}) => {
+    const currentTime = Date.now();
+    const filteredShifts = shiftDetails.filter(item => item.start_time < currentTime);
```
amalrzv, line 7: I like this filter, i think we might need to eventually incorporate some sort of edge case if the timezone differs unless they are sorting by location

client_app/src/components/shelter/PastVolunteers.js
```diff
@@ -1,22 +1,30 @@
...
 
     return (
-      <div className="volunteers-item-container">
-        <div className="volunteers-item">
-          <AccountCircleIcon />
-          <span>{shiftDetails.label}</span>
-          <NotificationsNoneIcon />
-        </div>
+      <div className="past-volunteers-container">
+        {uniqueWorker.length > 0 ? (uniqueWorker.slice(0,3).map((name, index) => (
```
amalrzv, line 14: Is this hardcoded just for now? If not we might need to expand this eventually ( where it limits the users that appear to 3) depending on how many volunteers a shelter requires.

tnhi26990: We only show three past volunteers on the Shelter Dashboard. Since this is a dashboard, we should only display a few.

client_app/src/components/shelter/PastVolunteers.js
tnhi26990: We use Unix epoch for the start_time and end_time, so it won't be affected by the user's time zone.

client_app/src/components/shelter/ShelterDashboard.js
```diff
@@ -3,8 +3,64 @@ import OpenRequest from "./OpenRequests";
...
+          fetch(request_endpoint, {
+            method: "GET",
+            headers: {
+              "Content-Type": "application/json",
+              Authorization: "volunteer@slu.edu",
+            },
+          })
+          .then((response) => response.json())
+          .then((data) => {
+            if (Array.isArray(data) && data.length > 0) {
```
amalrzv, line 46: just in case, add check if data doesnt exist or is null

tnhi26990: In this if statement, it checks two conditions at the same time: it needs to be an array, and its length is greater than 0, which means it cannot be empty.

Brijitha1609 (approved):
lgtm;

PR #153, opened 2024-11-25
Fixed: Added functionality to delete work shift from a volunteer with new database objects (Issue #150)

### LINKED ISSUE #150: Update Use Case for Deleting Shifts
Given the upcoming changes of the objects,
When updating the functionality for deleting shifts,
Then a new file should be created to accommodate the updated Volunteer object, while maintaining the current logic for shift deletions.

- **Create a new file:**
   Implement a new file called delete_volunteer_shift.py that handles the deletion request by volunteers for shifts they 
  have signed up for.

- **Maintain old logic:**
   The current delete_workshift.py file will not be deleted but left in place for any ongoing usage while the new logic is 
   implemented for the updated Volunteer object.

- **Adjustments for the new Volunteer object:**
    -Ensure the new functionality in delete_volunteer_shift.py reflects the use of UUIDs for the Volunteer object.
    -The same logic for shift deletions will be reused but modified to accommodate the changes in how Volunteer objects 
     are identified and managed.

**Acceptance Criteria**
- [ ] Creation of delete_volunteer_shift.py that mirrors the logic of delete_workshift.py while accounting for the new data objects.
- [ ] Ensure that UUIDs and any relevant changes in the new objects are properly handled in shift deletion requests.
- [ ] Test the new file to ensure correct functionality without impacting the existing shift deletion process.

cbiddle3:
Fixes #150 

**What was changed?**

Added the function to delete a shift from a Volunteer's (object) array of shifts, this is for the use case when a volunteer wants to cancel a shift they signed up for. 

**Why was it changed?**

To reflect the new structure for the back-end as it is updated to support shelter-side logic. 
Here are the objects defined in a doc: https://docs.google.com/document/d/193V1UcgqoHvLwOmyV8xwI5DB2p6P2sP92eAoi3Jg4go/edit?usp=sharing

**How was it changed?**

Created a new file "delete_volunteer_shift.py" that holds the logic to delete the shift instance and handles the error should a shift id not appear in the volunteer's array of signed_up_shifts. I added the functions that separate the Python MongoDB syntax in mongorepo.py and memrepo.py.

### DIFF
--- a/server/repository/memrepo.py
+++ b/server/repository/memrepo.py
@@ -43,7 +43,7 @@ class MemRepo:
                 return
 
     # This is for the new use case of a volunteer deleting a shift
-    def delete_shift_use_case(self, shift_id, user_id): 
+    def delete_shift_use_case(self, shift_id, user_id):
         for item in self.data:
             if shift_id in item["signed_up_shifts"] and item["id"] == user_id:
                 item["signed_up_shifts"].remove(shift_id)
--- a/server/repository/memrepo.py
+++ b/server/repository/memrepo.py
@@ -41,7 +41,7 @@ class MemRepo:
             if item["_id"] == shift_id:
                 self.data.remove(item)
                 return
-    
+
     def delete_shift_use_case(self, shift_id, user_id):
         for item in self.data:
             if shift_id in item["signed_up_shifts"] and item["id"] == user_id:
--- a/server/repository/memrepo.py
+++ b/server/repository/memrepo.py
@@ -41,4 +41,10 @@ class MemRepo:
             if item["_id"] == shift_id:
                 self.data.remove(item)
                 return
+    
+    def delete_shift_use_case(self, shift_id, user_id): 
+        for item in self.data:
+            if shift_id in item["signed_up_shifts"] and item["id"] == user_id:
+                item["signed_up_shifts"].remove(shift_id)
+                return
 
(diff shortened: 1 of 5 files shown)


server/repository/mongorepo.py
```diff
@@ -80,6 +80,15 @@ def delete(self, shift_id):
         self.collection.delete_one({"_id": ObjectId(shift_id)})
         return
 
+    def delete_volunteer_shift(self, shift_id, user_id):
```
Brijitha1609, line 83: Instead of adding this code for time being, you can just make comments about what should be done here.

server/repository/memrepo.py
```diff
@@ -42,3 +42,8 @@ def delete(self, shift_id):
                 self.data.remove(item)
                 return
 
+    def delete_shift_use_case(self, shift_id, user_id):
```
Brijitha1609, line 45: Instead of adding this code for time being, you can just make comments about what should be done here.

cbiddle3: This has to be here it's the function that uses Python to remove the shift, Nhi didn't have to do it since the function she needed already is made. I don't think we should make someone reimplement it later, as it belongs here and does not impact the other back-end functionality.

server/repository/mongorepo.py
cbiddle3: This has to be here it's the function that uses Python to remove the shift, Nhi didn't have to do it since the function she needed already is made. I don't think we should make someone reimplement it later, as it belongs here and does not impact the other back-end functionality.

Brijitha1609 (approved):
Good job! @cbiddle3.

PR #155, opened 2024-12-02
Change Object Name for the Add Shift Function

tnhi26990:
**What was changed?**

I changed the add_shift_volunteers function.

**Why was it changed?**

It was changed to be consistent with the new object.

**How was it changed?**

I imported the new object created and changed the object's name in the file. 


Brijitha1609 (approved):
lgtm;

PR #163, opened 2025-01-31
Send shelter-defined shifts to backend server

### LINKED ISSUE #160: Send Shelter-defined shifts to the backend server
**This task is related to the Shelter Dashboard Feature**
Shelter Dashboard currently only has client side application, with no backend support. Users of shelter dashboard can define shifts for when they need volunteers. This is done in the ```client_app/src/components/shelter/RequestForHelp.js```.  To see the details of what a shelter representative defines as part of a shift, run the application in development environment, sign in using developer@slu.edu email and any password and then point your browser to: http://localhost:3000/shelter-dashboard

From there, click on "Set Shifts" in the top navigation menu and you'll see a calendar view. Selecting some time slot in the calendar, pops up another window, where Shift Details are defined. When the user clicks "Submit Shift", the ```handleSaveEvent``` function of ```client_app/src/components/shelter/RequestForHelp.js``` is called, but the data doesn't really get saved and does not get sent anywhere.

When the user clicks on "Submit Shift", as described above, the data that the user populated as part of the shift details needs to be sent to the backend server. For now, the backend server needs to simply receive the data, and maybe print it to the terminal so that we can verify that data was received. Later, we will take that data and store it in the database. To make this happen, changes are needed on the server side and client_app side. 

On the client side, when user clicks "Submit Shift", call the server's /service_shift API endpoint (new endpoint to be added as part of this issue), sending the following data in JSON format:

- shelter_id (right now, shelter id is hard coded into the application in ShelterDashboard.js file.  We will need to figure out a way to pass shelter_id to the RequestForHelp.js module, but to keep things simple, you can just hard code this value for now.
- Shift start time, as defined by the user. This should be milliseconds since the Epoc in UTC
- Shift end time, as defined by the user. This should be milliseconds since the Epoc in UTC
- Number of volunteers required, as defined by the user
- Shift name, as defined by the user.

An example of a POST request going from the client to the backend server can be found in the submitShifts function of ```client_app/src/components/volunteer/Shelter.js```

On the server side, add an api endpoint /service_shift that handles a POST request. Other API endpoints exist in the ```server/application/rest/work_shift.py``` file and can be used as an example of how to create an API endpoint. This endpoint receives the request from the shelter dashboard,  extracts the JSON data, logs the data to console and sends a SUCCESS response, using the status code defined in HTTP_STATUS_CODES_MAPPING.ResponseTypes.SUCCESS (in the work_shift.py file).

**Success Criteria**

- [ ] When user clicks "Submit Shifts", as described in this issue report, user-defined shift data along with shelter ID (possibly hard coded) is sent to the backend server
- [ ] Backend server responds with SUCCESS and logs the received data to the terminal.

emrameduseljac:
Fixes #160 

**What was changed?**

Modified RequestForHelp.js, specifically updating the handleSaveEvent function to send shift data to the backend.
Added new service_shift endpoint in work_shift.py, to handle shift submissions and JSON parsing.
Added venv/ environment to .gitignore.

**Why was it changed?**

Previously, when shelter shifts were submitted, the information would go no where or be recognized at all. These changes ensure that when a shift is submitted, the data is sent to the backend and recognized for future changes on saving and displaying data.

**How was it changed?**

In RequestForHelp.js, updated handleSaveEvent to make a POST request to /service_shift.
Ensured the request sends as JSON format including the (hardcoded) shelter_id, shift_start, shift_end, and volunteers_required.

In work_shift.py, added /service_shift route to receive the shift data. Used request.get_json(force=True) to make sure JSON parsing occurs correctly.


client_app/src/components/shelter/RequestForHelp.js
```diff
@@ -5,6 +5,8 @@ import dayjs from "dayjs";
 import "react-big-calendar/lib/css/react-big-calendar.css";
 import { Modal, Button, Form } from "react-bootstrap";
 
+const SERVER = "http://localhost:5000";
```
kate-holdener, line 8: We have the SERVER defined in ```client_app/src/config.js``` file. You should use that definition instead of re-defining it here.

client_app/src/components/shelter/RequestForHelp.js
```diff
@@ -38,65 +40,35 @@ export default function RequestForHelp() {
...
-          selectedEvent &&
-          event.start === selectedEvent.start &&
-          event.end === selectedEvent.end
-        ) {
-          return false;
-        }
-    
-        const eventStart = new Date(event.start);
-        const eventEnd = new Date(event.end);
-        return newEvent.start < eventEnd && newEvent.end > eventStart;
```
kate-holdener, line 58: We probably don't want to remove these sanity checks on the user data.

client_app/src/components/shelter/RequestForHelp.js
```diff
@@ -38,65 +40,35 @@ export default function RequestForHelp() {
...
-      setVolunteerError("");
-      setShowModal(false)
-    } else if (!newEvent.volunteersRequired) {
-      setVolunteerError("'Number of Volunteers Required' field is numeric.");
-      return;
-    } else {
-      setVolunteerError("Required fields cannot be empty.");
-      return;
-    }
-
```
kate-holdener, line 98: These sanity checks should probably also stay. Is there a reason to remove them?

kate-holdener (changes requested):
@emrameduseljac Nice job on getting the data sent to the server! When I define a new shifts, I do see the data printed to the server console:
```
backend-1               | ==== PARSED JSON DATA ====
backend-1               | {
backend-1               |   "shelter_id": "12345",
backend-1               |   "shift_name": "",
backend-1               |   "shift_start": 1738416600000,
backend-1               |   "shift_end": 1738420200000,
backend-1               |   "volunteers_required": 0
backend-1               | }
backend-1               | 172.19.0.1 - - [31/Jan/2025 15:19:15] "POST /service_shift HTTP/1.1" 200 -
backend-1               | 172.19.0.1 - - [31/Jan/2025 15:19:45] "OPTIONS /service_shift HTTP/1.1" 200 -
backend-1               | ==== PARSED JSON DATA ====
backend-1               | {
backend-1               |   "shelter_id": "12345",
backend-1               |   "shift_name": "",
backend-1               |   "shift_start": 1738443600000,
backend-1               |   "shift_end": 1738441800000,
backend-1               |   "volunteers_required": 0
backend-1               | }
```
It looks like you removed some client side sanity checks on the shifts data (probably not intentionally). I left more detailed inline comments about that. Those will need to be addressed before we can merge this.

client_app/src/components/shelter/RequestForHelp.js
```diff
@@ -38,65 +40,35 @@ export default function RequestForHelp() {
...
+        setShowModal(false);
+      })
+      .catch((error) => {
+        console.error("Error submitting shift:", error);
+        setVolunteerError("Failed to submit shift.");
       });
-    
-      if (isOverlapping) {
-        setVolunteerError("This shift overlaps with an existing shift.");
-        return;
```
amalrzv, line 63: We had to add these checks in case that a volunteer selects multiple shifts at the same time (shows on the selected shifts section as overlapping or not overlapping). I would recommend keeping this.

server/application/rest/work_shift.py
```diff
@@ -253,3 +253,20 @@ def login():
...
     return result
+
+@blueprint.route("/service_shift", methods=["POST"])
+@cross_origin()
+def service_shift():
+    data = request.get_json(force=True)
+
+    if not data:
+        return jsonify({
+            "message": "Invalid or missing JSON"
```
amalrzv, line 264: Is there a way to specify whether it is just invalid or mission and have 2 separate messages?

kate-holdener: This is just a placeholder for now. We'll be replacing the backend logic with real processing of the data.

client_app/src/components/shelter/RequestForHelp.js
```diff
@@ -38,65 +39,65 @@ export default function RequestForHelp() {
...
-        setEvents((prev) =>
-          prev.map((ev) =>
-            ev.start === selectedEvent.start && ev.end === selectedEvent.end
-              ? {
-                  ...formattedEvent,
-                  title: formattedEvent.title,
-                }
-              : ev
-          )
-        );
```
kate-holdener, line 85: When I create a new event thorough the UI, I see that it gets sent to the server and event details are printed to the console on the server. However, the event does not show up in the calendar view of UI, like it used to. I think that this might be the code that made the newly created event show up.

kate-holdener (changes requested):
@emrameduseljac The functionality works as required. There is one "regression": the newly created event does not show in the calendar view of UI, while in the main branch, the newly created event shows up. I left an inline comment where I suspect the problem is.

Also, I noticed that when creating an event, the user can select end time to be before start time. However, this is not a regression, because the behavior is the same in the main branch. We can fix this as part of a different issue, unless you want to do a quick fix  for it here here.

client_app/src/components/shelter/RequestForHelp.js
kate-holdener: Looks like @emrameduseljac added the checks back in.

kate-holdener (approved):

PR #166, opened 2025-02-03
Struct rename

### LINKED ISSUE #159: Server side structure renaming
**This is a refactoring request**
When this project started, we anticipated that when volunteers sign up to work, they would define the start and end time of when they will be working (their work shift). As a result of this assumption, we defined a WorkShift class in the ```server/domains/work_shift.py`` file, with the necessary fields for user-defined work shift. This WorkShift would then get added to the volunteer's record in the database.

As the development went on, it was discovered that Shelters should be defining the work shifts and volunteers should be signing up for such shifts. The team defined new data structures to match this direction. Specifically,  the team defined a Work class in the ```server/domains/workshift.py''' file, with the necessary fields for defining shelter specified shifts.

Now we have classes that have similar names and confusing meaning. I suggest a small refactoring, to clean things up.

Since Shelters will be defining the time slots for when they need volunteers, let's call those time slots (along with the number of people needed) as ServiceShift. Essentially, the Work structure defined in ```server/domains/workshift.py''' should be renamed to ServiceShift and the file should be renamed to ```server/domains/service_shift.py'''. This should be a rename of a file (don't keep the old file). If there are dependencies on this file, those dependencies will need to be adjusted to use the new file and class name.

We will need to adjust how volunteers sign up for Service Shifts. When a volunteer signs up for a ServiceShift, a new object ServiceCommitment should be created. This object is defined by the volunteer id (a string) and the ServiceShift id (this should be defined in the ```server/domains/service_commitment.py''' file).  

Eventually, ServiceCommitment will replace what's currently a WorkShift (defined in ```server/domains/work_shift.py'''). However, we don't want to remove WorkShift class yet, because it's used by our API, which is used by the client side app. For now, park the WorkShift class as deprecated using the Deprecated python module. Any code that uses WorkShift should also be marked as depcrecated.

**Success Criteria**

- [ ] WorkShift class, defined in ```server/domains/work_shift.py''' is marked as deprecated
- [ ] Any function that uses WorkShift is marked as deprecated
- [ ] Work class in ```server/domains/workshift.py''' is renamed to ServiceShift, and the file is renamed to ```server/domains/service_shift.py. 
- [ ] Any function that uses Work class is adjusted to use ServiceShift class. 
- [ ] A new class defining the data model for shift sign up, ServiceCommitment, is defined in the server/domains directory. ServiceCommitment consists of a ServiceShift id (from the ServiceShift object) and a Volunteer ID (a string, usually an email address).

strejoarciles:
Fixes #159 
Now there is an import to deprecated which is reflected in the requirements.txt. Some other additions to requirements.txt include flask-cors, pymongo, and deprecated. One concern that I have is that these packages were already included in the requirements.txt, however my tests would not pass unless I had duplicated those packages. The work structure under server/domains/workshift.py was renamed to ServiceShift and the file was respectively renamed service_shift.py. Files that imported "from domains.work_shift import WorkShift" were changed to "from domains.service_shift import WorkShift" to reflect this change. This was done to reflect the changes in the dependencies within the files. The service_commitment.py was added to enable the eventual replacement of WorkShift.  I created a new file, server/domains/service_commitment.py, and defined the ServiceCommitment class. The ServiceCommitment class links a ServiceShift (via its ID) to a volunteer. 

The changes were made to address evolving requirements for how shelters and volunteers interact with work shifts. Initially, the application was designed under the assumption that volunteers would define their own shifts (start and end times). As development progressed, it became clear that shelters should instead define the shifts, specifying the time slots and the number of volunteers needed for each.

The Work class was renamed to ServiceShift, which better reflects its purpose as a shelter-defined shift. All references to the Work class were updated to use the new ServiceShift class, ensuring consistency and removing potential confusion in the codebase.

The code was changed by the renaming of files. Additionally, I went through and changed each time that WorkShift was imported to a file. Now, each file will reference the new files created (service_shift.py). 

The Work class was renamed to ServiceShift to reflect its role as the shelter-defined shift structure. This new class tracks volunteer sign-ups for shelter-defined shifts. It uses volunteer_id and ServiceShift ID as its fields, ensuring a clear and consistent data model. An important line that I added was "Deprecated>=1.2.10" to the requirements file. I also replaced the imports referring to WorkShift with from "domains.service_shift import WorkShift".

### DIFF
--- a/server/domains/service_shift.py
+++ b/server/domains/service_shift.py
@@ -9,8 +9,10 @@ import uuid
 import dataclasses
 from deprecated import deprecated
 
+@deprecated(
+    reason="WorkShift is deprecated and will be replaced by ServiceCommitment."
+)
 
-@deprecated(reason="WorkShift is deprecated and will be replaced by ServiceCommitment.")
 @dataclasses.dataclass
 class WorkShift:
     """
--- a/server/domains/service_shift.py
+++ b/server/domains/service_shift.py
@@ -1,16 +1,16 @@
 """
-This module defines data classes for service shifts, commitments, and a deprecated
-WorkShift class to support shelter service management.
+This module defines data classes for 
+service shifts and commitments,
+with the deprecated WorkShift 
+class marked for removal.
 """
 
 import uuid
 import dataclasses
-import warnings
 from deprecated import deprecated
 
-@deprecated(
-    reason="WorkShift is deprecated and will be replaced by ServiceCommitment."
-)
+
+@deprecated(reason="WorkShift is deprecated and will be replaced by ServiceCommitment.")
 @dataclasses.dataclass
 class WorkShift:
     """
@@ -20,7 +20,7 @@ class WorkShift:
     first_name: str
     last_name: str
     shelter: int
-    start_time: int  # Number of milliseconds since the Epoch in UTC
+    start_time: int
     end_time: int
     _id: uuid.UUID = None
 
@@ -35,19 +35,12 @@ class WorkShift:
     @classmethod
     def from_dict(cls, d):
         """Creates a WorkShift instance from a dictionary."""
-        warnings.warn(
-            "WorkShift is deprecated. Use ServiceCommitment instead.",
-            DeprecationWarning
-        )
         return cls(**d)
 
     def to_dict(self):
         """Converts the WorkShift instance to a dictionary."""
-        warnings.warn(
-            "WorkShift is deprecated. Use ServiceCommitment instead.",
-            DeprecationWarning
-        )
         return dataclasses.asdict(self)
+
 
 @dataclasses.dataclass
 class ServiceShift:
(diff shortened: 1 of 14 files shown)


server/domains/service_shift.py
```diff
@@ -0,0 +1,91 @@
...
+    service_shift_id: uuid.UUID
+
+    @classmethod
+    def from_dict(cls, d):
+        """Creates a ServiceCommitment instance from a dictionary."""
+        return cls(**d)
+
+    def to_dict(self):
+        """Converts the ServiceCommitment instance to a dictionary."""
+        return dataclasses.asdict(self)
```
kate-holdener, line 91: This class is defined in the service/comains/service_commitment.py. This is duplication.

server/domains/service_shift.py
```diff
@@ -0,0 +1,91 @@
...
+        self._id = new_id
+
+    @classmethod
+    def from_dict(cls, d):
+        """Creates a WorkShift instance from a dictionary."""
+        return cls(**d)
+
+    def to_dict(self):
+        """Converts the WorkShift instance to a dictionary."""
+        return dataclasses.asdict(self)
```
kate-holdener, line 44: This class should be defined in the file that you deleted (server/domains/work_shift.py)

kate-holdener:
Closing without merging: another PR was submitted to resolve the issue.

PR #170, opened 2025-02-09
Struct rename2

### LINKED ISSUE #159: Server side structure renaming

strejoarciles:
Fixes #159 

**What was changed?**
I properly renamed the correct file (workshift.py -> service_shift.py). I ensured that every file that referenced workshift was marked as deprecated as well as changed references to service_shift. I also created a unique ID for ServiceCommitment. 

**Why was it changed?**

To prepare for tying the shelter dashboard with volunteer dashboard: shelters will define 'service shifts' and volunteers will view these service shifts and create 'service commitment' by signing up for a shift.

**How was it changed?**
I added an entirely new file called service_commitment.py that contains the necessary ID's. I ensured that the code located in work_shift.py was marked as deprecated.


server/domains/service_shift.py
```diff
@@ -1,14 +1,14 @@
...
+This module handles data conversion from 
+dictionary to class obj or vice versa.
 """
 import uuid
 import dataclasses
 
-
 @dataclasses.dataclass
-class Work:
+class ServiceShift:  # renamed ServiceShift from Work
```
kate-holdener, line 9: Let's remove the comment, because it will be confusing in the future. 
```suggestion
class ServiceShift:
```

kate-holdener (changes requested):
@strejoarciles That's a good start. Here are some more changes that are needed:

- [ ] Any class that uses Work class should be modified to use ServiceShift. Here are some files that are still using Work class:
      - server/use_cases/add_volunteer_shifts.py (should be renamed to add_service_shifts.py and Work should be replaced with ServiceShift)
      - server/serializers/workshift.py (should be renamed to service_shift.py). The comments in this code are incorrectly referring to WorkShift object (this was a copy and paste error). Please change any WorkShift reference in this file to ServiceShift

- [ ] I see that you added `deprecated` package to the requirements.txt, but this package is not used in your changes. I'm fine with dropping the requirement to make WorkShift object deprecated. We can leave WorkShift as is, and once we finish the transition to ServiceShift and ServiceCommitment, we can simply remove all references to WorkShift. In that case, let's remove `deprecated` from requirements.txt

server/tests/use_cases/test_list_workshifts.py
kate-holdener, line 1: @strejoarciles you did not need to change this file. Please revert this change:
`git restore --source=5d14bdaf7d6a761 server/tests/use_cases/test_list_workshifts.py`

server/use_cases/add_service_shifts.py.py
kate-holdener, line 1: Your file name ends on py.py instead of one .py

server/serializers/service_shift.py
kate-holdener, line 1: Please change comments that reference WorkShift to reference ServiceShift

kate-holdener (changes requested):
@strejoarciles You have tests failing in the automated testing because you changed a file (test case) that did not need to be changed. Please see my additional inline comments.

kate-holdener:
@strejoarciles 
I see two options:
1.  Make WorkShift deprecated (as you did in your latest commit). Then, you do need to include deprecated in requirements.txt
2.  Don't make WorkShift deprecated. Then, undo your latest changes to the WorkShift class (where you marked it as deprecated).

The reason tests fail right now is because the system cannot find the 'deprecated' module (because it's not in requirements.txt)

kate-holdener (approved):

PR #162, opened 2025-01-30
login navigation

amalrzv:
Fixes #156 

**What was changed?**
Created HomeDashboard.js and HomeDashboard.css, changed App.js. HomeDashboard.js what created as a page that holds either role (volunteer or shelter admin) to then navigate to its respective login. App.js was modified to adjust routes for navigation. Created ShelterLogin.js based on Volunteer's Login.js  so that shelter admin can login once button is clicked on HomeDashboard.  Adjusted the route's in App.js accordingly. 

*Here, describe the issue that you are fixing with this code, and why your code fixes it.*
The codebase didn't have a login for shelter admin, which was added. Additionally, the app wasn't routed to the new homedashboard, so once that was created along with its respective styling, that was routed. 

**How was it changed?**

*Here, get into detail about what files you modified, and talk about the most important lines in regards to fixing the issue.*

**Screenshots that show the changes (if applicable):**
Screenshot 2025-02-03 at 1 42 44 PM
Screenshot 2025-02-03 at 1 43 01 PM

PR #171, opened 2025-02-10
Fixed home dashboard

kate-holdener:
When we merged PR #162, we introduced a small problem. If the user already had an authentication token stored in browser cache, no dashboard would show up. This had to do with the way HomeDashboard was defined: it was attempting to re-direct the user to either VolunteerDashboard or ShelterDashboard based on the token that's stored in browser cache. However, for now our token does not distinguish between volunteer and shelter admin, so no dashboard would show up instead.

**What was changed?**
Modified HomeDashboard and App in the client_app, by removing various routes that attempted to do a smart redirect.

### DIFF
--- a/client_app/src/App.js
+++ b/client_app/src/App.js
@@ -15,7 +15,7 @@ import { ShiftDetails } from "./components/shelter/ShiftDetails";
 import UpcomingRequests from "./components/shelter/UpcomingRequests";
 import "./styles/App.css";
 import HomeDashboard from "./components/HomeDashboard";
-import { BrowserRouter as Navigate, useLocation } from "react-router-dom";
+import { useLocation } from "react-router-dom";
 
 function NavigationControl({ auth }) {
   const location = useLocation(); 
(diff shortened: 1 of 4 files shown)


client_app/src/components/authentication/Login.js
```diff
@@ -39,7 +39,7 @@ export default function Login({ setAuth, userRole }) {
 
   const handleSubmit = async (e) => {
     e.preventDefault();
-    await LoginUser(userRole, username, password);
```
amalrzv, line 42: I like how this was implemented without userRole, i realize looking now it was redundant/unnecessary

client_app/src/components/HomeDashboard.js
```diff
@@ -1,18 +1,9 @@
...
 function HomeDashboard() {
-    const navigate = useNavigate(); 
-    useEffect(() => {
-      const token = localStorage.getItem("token");
-        
-      if (token) {
-        navigate("/home-dashboard");
-      }
-    }, [navigate]);
-
```
amalrzv, line 15: From my understanding, this was removed because we don't have a way to determine if the token stored is an authenticated volunteer or an authenticated shelter staff member. Thank you for making this change!

client_app/src/App.js
```diff
@@ -54,16 +54,6 @@ function App() {
...
-            <Route
-              path="/home-dashboard"
-              element={
-                auth ? (
-                  <Navigate to="/volunteer-dashboard" />
-                ) : (
-                  <Navigate to="/shelter-dashboard" />
-                )
-              }
-            />
```
amalrzv, line 66: I ran it and it works great on my end! Just curious, was this removed because of the same idea with homedashboard.js? Or because it wasn't necessary for navigation?

amalrzv (commented):
I ran the code and it looks perfect on my end. Thank you for making these changes!

client_app/src/App.js
kate-holdener: I was seeing behavior that when I go to the main page, I would get automatically redirected to /home-dashboard (I already had the token), but the screen would show up as blank. I am not 100% sure why it was happening, so I was trying to simplify the routing that's happening.

amalrzv (approved):

PR #173, opened 2025-02-12
Issue 104

### LINKED ISSUE #104: Change UI testing to use mocking instead of making API and system calls
UI testing currently makes API calls to GetHelp and uses real-system time. These should be mocked instead to avoid tests breaking. 

**Describe alternatives you've considered**
- MSW (mock service worker) for GetHelp API calls
- Jest mocking feature to set the system time instead of using real-time.
- Set up MSW for UI testing that involves the server.

kate-holdener:
Fixes #104

**What was changed?**

In the client_app, the VolunteerDashboardShelterCards test would intermittently fail, because it relied on a call to GetHelp API, and had an await function with a timeout. When the response was slower than the expected timeout, the test would fail. 

**Why was it changed?**

We noticed that VolunteerDashboardShelterCards would occasionally fail in GitHub actions, and had a temporary solution by increasing the timeout interval. This was not a stable solution.

**How was it changed?**

The test code was modified to not call the real api but to mock out the shelter data, removing the await and the timeout. In order to make this happen, I refactored the `client_app/src/components/volunteer/Shelters.js` file, moving out the code for retrieving shelter information from GetHelp into `client_app/src/components/volunteer/hooks/useShelterData.js`.


kate-holdener:
Searched for other places in the client_app that are making calls to GETHELP_API and did not find any. The only location was in Shelters.js (this call is now in hooks/useShelterData.js, so it can be mocked during tests).

PR #178, opened 2025-02-17
Signing up for predefined shifts

### LINKED ISSUE #168: Volunteer Dashboard: sign up for shifts should show shelter-defined shifts
Currently, when a volunteer signs up for shifts, they pick any start and end time they want. In reality, shelters will be defining shifts for when they need volunteers. Volunteers should select their shifts from the shelter-defined options. This feature request is for client side changes only. 

When issue #165 is resolved, we will be able to retrieve shelter-defined shift data to the volunteer dashboard. For now, we can define JSON data in the volunteer dashboard with some hard-coded start and end times, and use that data as shelter-defined shifts. 

Display shelter-defined shifts as available options to the user, instead of allowing the user to arbitrarily select start and end time. This is a client-side change only.

**Success Criteria**
- [ ] The user should be able to sign up for one or multiple shelter-defined shifts. 
- [ ] The user should not be able to arbitrarily select start and end time for when they want to work.

emrameduseljac:
Fixes #168 

**What was changed?**

When you browse the "Sign up for shifts" page, it will display hardcoded predefined shifts from the various shelters, which the user can add to their cart and submit. Users are now unable to add their own shifts, they can only select from the provided shifts.

**Why was it changed?**

These changes were made so volunteers can select from given shifts, so they know what they are signing up for, and shelters can decide ahead of time what shifts they would need.

**How was it changed?**

Changes were specifically made to `Shelters.js`. In Shelters.js, there are now hardcoded predefined shifts that should be used later on. The `manageShifts` function now makes sure that the shifts start and end in UTC to avoid any time zone issues. There is also a unique ID identifier now, to avoid any duplicates. `fetchData` now takes from the predefined shifts.

**Current status:**

The predefined shifts are not being shown on the website. Times are still adjustable to the volunteer user. One of the tests do not pass.

### DIFF
--- a/client_app/src/components/volunteer/Shelters.js
+++ b/client_app/src/components/volunteer/Shelters.js
@@ -2,7 +2,7 @@ import { useState, useEffect } from "react";
 import ShelterList from "./ShelterList";
 import ConfirmationPage from "./ConfirmationPage";
 import { Pagination } from "./Pagination";
-import { SERVER } from "../../config";
+import { GETHELP_API, SERVER } from "../../config";
 import { Link } from "react-router-dom";
 import getAuthHeader from "../../authentication/getAuthHeader";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
@@ -17,9 +17,9 @@ const Shelters = (props) => {
   let defaultRadius = "5";
   if (props.condensed) defaultRadius = "25";
   const [data, setData] = useState([]);
-  const [setLatitude] = useState(33.997103);
-  const [setLongitude] = useState(-118.4472731);
-  const [setRadius] = useState(defaultRadius);
+  const [latitude, setLatitude] = useState(33.997103);
+  const [longitude, setLongitude] = useState(-118.4472731);
+  const [radius, setRadius] = useState(defaultRadius);
   const [loading, setLoading] = useState(true);
   const [isButtonDisabled, setButtonDisabled] = useState(true);
   const [selectedShifts, setSelectedShifts] = useState([]);
@@ -37,6 +37,30 @@ const Shelters = (props) => {
   const shakeAnimation = useSpring({
     transform: shaking ? "translateY(-20px)" : "translateY(0px)",
   });
+
+  useEffect(() => {
+   fetchData();
+  }, [latitude, longitude, radius]);
+
+  const fetchData = () => {
+    setLoading(true);
+    let newEndpoint =
+      GETHELP_API +
+      "v2/facilities?page=0&pageSize=1000&latitude=" +
+      latitude +
+      "&longitude=" +
+      longitude +
+      "&radius=" +
+      radius;
+    fetch(newEndpoint, {
+      method: "GET",
+      headers: {
+        "Content-Type": "application/json",
+      },
+    })
+      .then((response) => response.json())
+      .catch((error) => console.log(error));
+  };
 
   useEffect(() => {
     // convert ShiftsData into an array format expected by ShelterList
--- a/client_app/src/components/volunteer/Shelters.js
+++ b/client_app/src/components/volunteer/Shelters.js
@@ -2,7 +2,7 @@ import { useState, useEffect } from "react";
 import ShelterList from "./ShelterList";
 import ConfirmationPage from "./ConfirmationPage";
 import { Pagination } from "./Pagination";
-import { GETHELP_API, SERVER } from "../../config";
+import { SERVER } from "../../config";
 import { Link } from "react-router-dom";
 import getAuthHeader from "../../authentication/getAuthHeader";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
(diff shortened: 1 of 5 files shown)


client_app/src/components/volunteer/Shelters.js
```diff
@@ -129,25 +146,26 @@ const Shelters = (props) => {
   function manageShifts(shift) {
     setShaking(true);
     setTimeout(() => setShaking(false), 200);
-    //setSelectedShifts([...selectedShifts, shift]);
-    //setButtonDisabled(selectedShifts.length === 0);
-    const shiftStart = dayjs(shift.start);
-    const shiftEnd = dayjs(shift.end);
+
+    const shiftStart = dayjs(shift.start).utc();
+    const shiftEnd = dayjs(shift.end).utc();
```
amalrzv, line 151: I'll have to look at the other file but the .utc() might be redundant because the .start and .end retrieves the .utc(). Maybe .start_time would also work since its gives it in MMM, dd, yyyy, HH:mm aa.

client_app/src/components/volunteer/Shelters.js
```diff
@@ -12,6 +12,19 @@ import { useSpring, animated } from "@react-spring/web";
...
+  "shelter-1": [
+    { id: 1, start: "2025-02-17T09:00:00", end: "2025-02-17T13:00:00", title: "Morning Shift" },
+    { id: 2, start: "2025-02-17T13:00:00", end: "2025-02-17T17:00:00", title: "Afternoon Shift" },
+    { id: 3, start: "2025-02-17T17:00:00", end: "2025-02-17T21:00:00", title: "Evening Shift" }
+  ],
+  "shelter-2": [
+    { id: 4, start: "2025-02-17T08:00:00", end: "2025-02-17T12:00:00", title: "Early Shift" },
+    { id: 5, start: "2025-02-17T12:00:00", end: "2025-02-17T16:00:00", title: "Mid-day Shift" }
+  ]
+};
```
amalrzv, line 26: I think based on the shifts now, I think that we can get rid of a lot of the labeling, which is done in a different file I believe. I think that Early shift and morning shift  with the afternoon shift overlaps with it and the mid-day shift can be kind of confusing because the times overlap. Is there a specific reason that these are labeled differently or is it labeled based on what the shelter wants?

client_app/src/components/volunteer/Shelters.js
```diff
@@ -104,7 +117,11 @@ const Shelters = (props) => {
     })
       .then((response) => response.json())
       .then((data) => {
-        setOriginalData(data.content);
+        const enhancedData = data.content.map(shelter => ({
+          ...shelter,
+          predefinedShifts: PREDEFINED_SHIFTS[`shelter-${shelter.id}`] || []
```
amalrzv, line 122: should this be =$ instead?

client_app/src/components/volunteer/Shelters.js
```diff
@@ -12,6 +12,19 @@ import { useSpring, animated } from "@react-spring/web";
...
+// hardcoded predefined shifts
+const PREDEFINED_SHIFTS = {
+  "shelter-1": [
+    { id: 1, start: "2025-02-17T09:00:00", end: "2025-02-17T13:00:00", title: "Morning Shift" },
+    { id: 2, start: "2025-02-17T13:00:00", end: "2025-02-17T17:00:00", title: "Afternoon Shift" },
+    { id: 3, start: "2025-02-17T17:00:00", end: "2025-02-17T21:00:00", title: "Evening Shift" }
+  ],
+  "shelter-2": [
+    { id: 4, start: "2025-02-17T08:00:00", end: "2025-02-17T12:00:00", title: "Early Shift" },
+    { id: 5, start: "2025-02-17T12:00:00", end: "2025-02-17T16:00:00", title: "Mid-day Shift" }
```
kate-holdener, line 24: The timestamps should be in milliseconds since the Epoc in UTC (that's what the server will be sending you eventually)

kate-holdener (changes requested):
I think you'll need to be changing IndividualShelter.js component. That's where we currently have the functionality allowing the users to select the dates and times of their shifts. I am not seeing your hard-coded data show up in the UI in the "Sign up for Shifts" page.

client_app/src/components/volunteer/Shelters.js
kate-holdener: I recommend placing the hard-coded data into its own file, and import it in the IndividualShelter.js

client_app/src/components/volunteer/ShiftsData.js
```diff
@@ -0,0 +1,42 @@
+const ShiftsData = {
+  "shelter-1": {
```
kate-holdener, line 2: This should be a shelter id, for example 6093. I bet this would fix your test problem.

client_app/src/components/volunteer/Shelters.js
```diff
@@ -11,14 +11,15 @@ import { faCalendarDays, faArrowRight, faCircleXmark } from "@fortawesome/free-s
...
 const Shelters = (props) => {
   let defaultRadius = "5";
   if (props.condensed) defaultRadius = "25";
   const [data, setData] = useState([]);
-  const [latitude, setLatitude] = useState(33.997103);
-  const [longitude, setLongitude] = useState(-118.4472731);
-  const [radius, setRadius] = useState(defaultRadius);
+  const [setLatitude] = useState(33.997103);
+  const [setLongitude] = useState(-118.4472731);
+  const [setRadius] = useState(defaultRadius);
```
kate-holdener, line 22: This change is causing problems, when I click on "show opportunities near me". You should undo this change.
```
Uncaught runtime errors:
×
ERROR
setLatitude is not a function
TypeError: setLatitude is not a function
    at setLocation (http://localhost:3000/static/js/bundle.js:7957:5)
ERROR
setLatitude is not a function
TypeError: setLatitude is not a function
    at setLocation (http://localhost:3000/static/js/bundle.js:7957:5)
```

client_app/src/components/volunteer/Shelters.js
```diff
@@ -38,8 +39,19 @@ const Shelters = (props) => {
   });
 
   useEffect(() => {
-    fetchData();
-  }, [latitude, longitude, radius]);
```
kate-holdener, line 42: You should undo this change as well.

kate-holdener (changes requested):
@emrameduseljac 
I posted a couple of comments for changes that should be undone. Also, it would be nicer if the time slots had a visual gap between them, as in the prototype below.
Screenshot 2025-02-24 at 12 03 42 PM

kate-holdener (approved):
Looks great!

PR #206, opened 2025-03-03
Issue 175

### LINKED ISSUE #175: Filter service commitments by service shift id
When issue #164  is resolved, we will have a /service_commitment API endpoint, allowing a volunteer to POST and GET their service commitments. 

Shelters will need to retrieve service commitments to know which and how many volunteers signed up to work for that shelter. To support this, the GET /service_commitment API endpoint need to support retrieval of all commitments for a given service shift.

Add a filtering parameter to the GET /service_commitment API endpoint, to allow specifying the shift_id we are interested in. When a GET request is sent to `/service_commitment?shift_id=<SHIFT_ID>`, the API should return only commitments for the given service shift id.

**Additional context**
We will later need to expand on this functionality and add time filtering, allowing shelters to retrieve only future commitments or only past commitments. We will also eventually need to address the authentication element of this: only users that are authorized admins for a given shelter should be able to retrieve commitments for that shelter. But we are saving this for the future, to keep the scope of this issue small.

**Success criteria**

- [ ] All available service commitments for a given shift id are returned with a GET request to `/service_commitement?shift_id=<SHIFT_ID>`
- [ ] Only service commitments for the given shift_id are returned (service commitments for other shifts are not included in the results)
- [ ] If there are no service commitments for a given shift_id, then API endpoint responds with 'success' and contains an empty list for results.

To populate service commitments, use POST requests to /service_commitment API endpoint, as described in issue #164

strejoarciles:
Fixes #175 

I made adjustments to resolve the parameter name inconsistency in the service commitment API. The issue occurred because our endpoint was using "shift_id" as the URL query parameter, while our internal data model used "service_shift_id". This mismatch caused filtering by shift ID to fail, returning empty results even when commitments existed. I modified the fetch_service_commitments function in service_commitment.py to use "service_shift_id" consistently throughout, including the parameter extraction from the request, conditional logic for user filtering, and the parameter passed to the list_service_commitments function. This change ensures that requests using the correct parameter name (/service_commitment?service_shift_id=13435) will properly filter and return all commitments associated with the specified shift.


Taktar (commented):
The changes appear to be focused on improving code readability and consistency. Nice job on making the codebase cleaner!

Taktar:
Issue:
import { render, screen, waitFor } from "@testing-library/react";
waitFor is imported but never used.

Suggestion:
Remove the unused import to clean up the code:
import { render, screen } from "@testing-library/react";

server/repository/mongod/service_commitments.py
kate-holdener, line 1: Looks like you changed mongo repository to mongod. Why was this needed?

server/config/mongodb_config.py
```diff
@@ -23,7 +23,7 @@ def load_env_file():
 class MongoConfig(object):
     """Base configuration class."""
     load_env_file()
-    MONGODB_HOST = os.getenv('MONGODB_HOST', 'mongodb')
+    MONGODB_HOST = os.getenv('MONGODB_HOST', 'mongod')
```
kate-holdener, line 26: When connecting to Mongodb Atlas, the host needs to be 'mongodb'. You should undo this change.

server/application/rest/service_commitment.py
```diff
@@ -76,9 +76,12 @@ def create_service_commitment():
 @service_commitment_bp.route("/service_commitment", methods=["GET"])
 def fetch_service_commitments():
     """
-    Handle GET request to retrieve service commitments for a user.
+    Handle GET request to retrieve service commitments.
+    Can filter by user (from token) and optionally by shift_id.
     """
     try:
+        # Extract shift_id from query parameters if provided
+        shift_id = request.args.get("shift_id")
```
kate-holdener, line 84: When I create two commitments with:
```
curl --location '127.0.0.1:5001/service_commitment' \
--data '[{"service_shift_id": "134134"}, {"service_shift_id": "13435"}]'
```
And then try to retrieve service commitments by shift id 13435, I am getting an empty list. The reason for that is inconsistency in the parameter we are passing to the request: /service_commitment?**shift_id**, while our internal field is called **service_shift_id**. I think the easiest way to address this is to make the URL parameter consistent with our internal field, and change the request to: /service_commitment?**service_shift_id**

Also note that when creating service shifts, we have to pass a string value for the service_shift_id (not an integer, as I originally wrote in the issue description).

kate-holdener, line 84:
```suggestion
        shift_id = request.args.get("service_shift_id")
```

server/.env.development
kate-holdener, line 1: Looks like you accidentally checked in this file

kate-holdener (changes requested):
@strejoarciles Nice job on making progress with this. You are really close. A couple if things to fix.

1. You added some unnecessary changes:
    - renamed repository/mongo directory to repository/mongod, which resulted in other changes (to import statements) that didn't need to happen
    - committed .env.development file
    - unnecessary change in config/mongodb_config.py file
 2. The filtering doesn't work because of an inconsistency between our URL parameter and internal structure. I provided a detailed inline comment for this with an easy solution.

kate-holdener (approved):

PR #219, opened 2025-03-07
Created tests for issue #189

### LINKED ISSUE #189: Server side allows to create service commitments for non-existing shifts
**Describe the bug**
The `add_service_commitments(commitments_repo, shifts_repo, commitments)` function in use_cases/add_service_commitments.py accepts commitments for non-existing service shifts. 

Specifically, the commitments parameter is a list of ServiceCommitment objects. Each ServiceCommitment object has an service_shift_id, representing a ServiceShift object. However, we never verify that the service_shift_id in the ServiceCommitment object corresponds to a real ServiceShift in our system.

**To Reproduce**
1. Start the server connected to a database
2. Send a POST request to /service_commitment endpoint with any service shift id. For example:
```
--data '[{"service_shift_id": "anything-you-want"}]'
```
3. Send a GET request to /service_commitment endpoint and verify that the commitment got erroneously created:
```
--data ''
```

**Expected behavior**
The system should only allow to create service commitments with valid shift ids.

**Suggested approach**
Add tests to server/tests/use_cases/add_service_commitments.py file testing valid and invalid scenarios.
- Test adding a commitment for a valid shift id -> test should pass
- Test adding a commitment for an invalid shift id -> test should vail

Update the `add_service_commitments(commitments_repo, shifts_repo, commitments)` function to ensure that ServiceCommitments we are creating have real existing shifts in the shifts_repo. If for a given ServiceCommitment, the ServiceShift does not exist, do not create that commitment in the commitments_repo. In the response for that service commitment, use 'success': false and 'message': 'cannot commit to non-existing shift <shift_id>'. Note that the commitments list may have commitments with valid shift ids, which should get created successfully.

After successfully updating add_service_commitments function, tests that were failing earlier should pass.

**Success criteria**

- [ ] A POST request to create one commitment for non-existing shift returns success with a data 
```
[
  {
     'service_commitment_id': None,
     'success': false, 
     'message': 'cannot commit to non-existing shift SHIFT_ID'
  }
]
```
Here, SHIFT_ID is replaced with the id of the shift that was targeted in the commitment
- [ ] A POST request to create multiple commitments: one for existing shift id and one for non-existing shift id returns success with data:
```
[
  {
     'service_commitment_id': "some real id",
     'success': true, 
  },
  {
     'service_commitment_id': None,
     'success': false, 
     'message': 'cannot commit to non-existing shift SHIFT_ID'
  }
]
```
- [ ] Test cases for valid and invalid shift ids should pass.

rohitparapati:
Fixes #189 

**What was changed?**

Added unit tests for the add_service_commitments function in server/tests/use_cases/service_commitments/test_add_commitments.py.
Mocked commitments_repo and shifts_repo using MagicMock() to isolate function behavior.

**Why was it changed?**

Ensures commitments are only created for existing shifts.
Supports regression testing for issue #189 to prevent creating commitments for non-existing shifts.

**How was it changed?**

Added three test cases:
Existing shift → Commitment should be created.
Non-existing shift → Commitment should not be created.
Mixed case → First and third commitments succeed, second (invalid) fails.
Ensured Pylint compliance by fixing long lines and removing unused imports.

### DIFF
--- a/server/tests/use_cases/service_commitments/test_add_commitments.py
+++ b/server/tests/use_cases/service_commitments/test_add_commitments.py
@@ -2,10 +2,18 @@
 This module contains tests for the service_commitments use case, focusing on 
 adding service commitments.
 """
-import pytest
-from unittest.mock import MagicMock
 import sys
 import os
+from unittest.mock import MagicMock
+import pytest
+
+# Ensure correct import path for the project
+sys.path.append(
+    os.path.abspath(
+        os.path.join(os.path.dirname(__file__), "../../../..")
+    )
+)
+
 from server.use_cases.add_service_commitments import add_service_commitments
 
 # Ensure correct import path for the project
@@ -34,7 +42,7 @@ def test_add_service_commitment_non_existing_shift(mocked_repos):
     # Mock commitment object
     commitment = MagicMock()
     commitment.service_shift_id = "non_existing_shift"
-    commitment.to_dict.return_value = {"_id": "None"} 
+    commitment.to_dict.return_value = {"_id": "None"}
 
     # Call function
     result = add_service_commitments(
@@ -48,4 +56,3 @@ def test_add_service_commitment_non_existing_shift(mocked_repos):
     commitments_repo.insert_service_commitments.assert_called_once_with(
         [{"_id": "None"}]
     )
-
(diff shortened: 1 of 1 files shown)


kate-holdener:
All these test cases pass on our current code base, however, our code currently does not prohibit creating service commitments for non-existing shifts.

PR #221, opened 2025-03-07
Added Server Side Tests

### LINKED ISSUE #212: Add data validation to the server side
We have built several API end points on the server. Each endpoint creates a relevant domain object (Shelter, ServiceShift, ServiceCommitment, etc). These domain objects are then passed to the relevant use cases. We need to add data validation on different levels:
1. API level - validate inputs received in the request, and return an error message if any required inputs are missing
2. Domain object level - add validation in the constructor/factory, to ensure domain objects have required and sensible data fields
3. Use case level - add use case specific validation.

smallrussian:
Fixes #212

What was changed?

    Added a new test file: server/tests/rest/service_shift/test_service_shift.py, which contains unit tests for the /service_shift REST API endpoint.
    Updated the import paths and patch decorators within the test file to ensure proper resolution of modules when running tests from different directories.

Why was it changed?

    Issue #212 highlighted that there were no test cases for the /service_shift endpoint.
    This change ensures that both the GET and POST methods for the endpoint are covered by automated tests. It protects against regressions by verifying that the endpoint returns the correct responses when interacting with mocked business logic components.

How was it changed?

    Test File Creation:
    A new file, server/tests/rest/service_shift/test_service_shift.py, was created. This file contains two main tests:
        A POST test that sends a JSON payload with proper headers, mocks the shift_add_use_case, and asserts that the endpoint returns a 200 status code along with the expected JSON response.
        A GET test that mocks the service_shifts_list_use_case and asserts that the endpoint returns a predefined list of shifts.

    Import and Patch Updates:
    To resolve module import issues, the import statements in the test file were updated to use relative paths (e.g., from application.rest.service_shifts import app), and the patch decorators were adjusted accordingly (e.g., @patch('application.rest.service_shifts.shift_add_use_case')). This ensures the tests run correctly both locally and in GitHub Actions.

    Testing Framework:
    The tests leverage Python’s unittest framework and use unittest.mock.patch to isolate the REST API layer from its underlying business logic, ensuring that only the API behavior is verified.

### DIFF
--- a/server/tests/rest/service_shift/test_service_shift.py
+++ b/server/tests/rest/service_shift/test_service_shift.py
@@ -1,14 +1,14 @@
 import unittest
 import json
 from unittest.mock import patch
-from server.application.rest.service_shifts import app
+from application.rest.service_shifts import app  # Updated import
 
 class TestServiceShiftAPI(unittest.TestCase):
     def setUp(self):
         self.client = app.test_client()
         self.client.testing = True
 
-    @patch('server.application.rest.service_shifts.shift_add_use_case')
+    @patch('application.rest.service_shifts.shift_add_use_case')
     def test_post_service_shift(self, mock_shift_add_use_case):
         # Arrange: Set up the expected IDs returned by the use case.
         expected_ids = [101, 102]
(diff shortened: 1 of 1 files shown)


kate-holdener (approved):
Nice job!

PR #236, opened 2025-03-24
progress on ShelterID Datatype fix

### LINKED ISSUE #230: Incorrect datatype for shelter_id in ServiceShift
**Describe the bug**
We originally defined ServiceShift in server/domains/shervice_shift.py to have an int for shelter_id. This is because shelter ids provided by GetHelp were integers. When we added shelters to our server, the ID that gets assigned to each shelter is a string (generated by mongodb). This means we now have an inconsistency: our shelter ids are strings, but we expect a shelter id as an integer on the server side. This inconsistency needs to be address.

**To Reproduce**
1. Create a service_shift through the /service_shift API endpoint
2. Examine the data in your mongodb - you will see that shelter_id is an integer

**Expected behavior**
The shelter_id associated with service shifts needs to be a string. The modules that process service shifts need to be updated accordingly. These modules include:
- /service_shift api endpoint
- ServiceShift domain object
- service_shift serializer
- ServiceShiftMongoRepo

**Suggested approach**
Update test cases in `server/tests/rest/service_shift/test_service_shift.py` to send an alpha-numeric string as a shelter_id, when making a POST request to /sevice_shift API. Run these tests with pytest and see them fail. Then updated the necessary server side code elements that expect shelter_id to be an integer (make them expect shelter_id to be a string). After successfully updating these code elements, all tests should pass.

Add a new test case to `server/tests/rest/service_shift/test_service_shift.py`, testing the following scenarios:
- add a service shift with shelter_id as "ID1"
- issue a GET request to /service_sift?shelter_id=ID1
- The response should have the service shift you created earlier.

Note that initially, this test case should fail, until you successfully fix the bug described in this issue.

**Success Criteria** 
1. A user must be able to create a service shift through the /service_shift API endpoint and the shelter_id that gets stored in the database must be a string. A user should also be able to call a GET /service_shift API endpoint, filtering the shifts by shelter ID (this functionality is already supported, just needs to be updated to account for shelter id being a string).

Specifically, the following scenario should work:
- A user creates a shelter (either through the admin dashboard or with a POST request as 
  ```
  curl --location '127.0.0.1:5000/shelter' \
  --data '{"name": "SLU winter inn", "address": {"street1": "1 Grand Ave", "street2": "", "city":"St. Louis", "state": "MO", "postalCode": "12345", "coordinates": {"latitude": 90, "longitude":-30}}}'
     ```
- A user creates a service sift for this shelter's shelter_id. Use a POST request to create a service shift:
```
--data '[
    {"shelter_id":"YOUR_SHELTER_ID", "shift_start": 10, "shift_end": 20}
]'
```
- A user retrieves all service shifts for the given shelter with:
```
curl --location '127.0.0.1:5000/service_shift?shelter_id=YOUR_SHELTER_ID' \
--data ''
```
   This should result in the service shift you created earlier.

2. Tests in `server/tests/rest/service_shift/test_service_shift.py` must be updated as described in this issue and they must pass.

amalrzv:
Fixes #230 

**What was changed?**
Modified server/domains/service_shift.py and server/tests/rest/service_shift/test_service_shift.py. The new test was added to verify that the system can filter and retrieve service shifts using string shelter IDs. It confirms the process of creating a shift with a string ID and then successfully filtering shifts by that string ID in a GET request. This should fail until everything is resolved.

### DIFF
--- a/server/domains/service_shift.py
+++ b/server/domains/service_shift.py
@@ -11,7 +11,7 @@ class ServiceShift:
     """
     Data class for work shift-related data.
     """
-    shelter_id: int
+    shelter_id: str
     shift_start: int  # Number of milliseconds since the Epoch in UTC
     shift_end: int
     shift_name: str = 'Default Shift'
--- a/server/tests/rest/service_shift/test_service_shift.py
+++ b/server/tests/rest/service_shift/test_service_shift.py
@@ -81,7 +81,7 @@ class TestServiceShiftAPI(unittest.TestCase):
                 "shift_name": "Default Shift"
             }
         ]
-        mock_list_use_case.return_value = [ServiceShift.from_dict(shift) for shift in expected_shifts]
+        mock_list_use_case.return_value = [ServiceShift.from_dict(expected_shifts)]
 
         # Act: Send GET request.
         response = self.client.get("/service_shift")
--- a/server/tests/rest/service_shift/test_service_shift.py
+++ b/server/tests/rest/service_shift/test_service_shift.py
@@ -81,7 +81,7 @@ class TestServiceShiftAPI(unittest.TestCase):
                 "shift_name": "Default Shift"
             }
         ]
-        mock_list_use_case.return_value = [ServiceShift.from_dict(shift) 
+        mock_list_use_case.return_value = [ServiceShift.from_dict(shift)
                                            for shift in expected_shifts]
 
         # Act: Send GET request.
--- a/server/tests/rest/service_shift/test_service_shift.py
+++ b/server/tests/rest/service_shift/test_service_shift.py
@@ -81,7 +81,8 @@ class TestServiceShiftAPI(unittest.TestCase):
                 "shift_name": "Default Shift"
             }
         ]
-        mock_list_use_case.return_value = [ServiceShift.from_dict(shift) for shift in expected_shifts]
+        mock_list_use_case.return_value = [ServiceShift.from_dict(shift) 
+                                           for shift in expected_shifts]
 
         # Act: Send GET request.
         response = self.client.get("/service_shift")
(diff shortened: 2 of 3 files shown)


server/domains/service_shift.py
kate-holdener, line 1: I can create a shelter and a service shift for that shelter's id. When I the GET /service_shift?shelter_id=MY_SHELTER_ID, I get no results. This is because the GET /service_shift endpoint tries to convert shelter_id to an integer (instead of a string).

server/tests/rest/service_shift/test_service_shift.py
```diff
@@ -98,5 +98,40 @@ def test_get_service_shift(self, mock_list_use_case):
...
+          "required_volunteer_count": 1,
+          "max_volunteer_count": 5,
+          "can_sign_up": True,
+          "shift_name": "Default Shift"
+        }
+        mock_list_use_case.return_value = [expected_shift]
+
+        # Act: Send GET request with shelter_id filter
+        response = self.client.get(
+            f"/service_shift?shelter_id={test_shelter_id}")
```
kate-holdener, line 119: the GET /service_shift API endpoint still converts shelter_id that's passed to an int, which is probably why this test case fails.

kate-holdener (changes requested):
@amalrzv the retrieval of service shifts filtered by ID still doesn't work, probably because GET /service_shift api endpoint converts shelter_id to an int, but we are passing values that cannot be converted to an integer.

server/domains/service_shift.py
```diff
@@ -11,7 +11,7 @@ class ServiceShift:
     """
     Data class for work shift-related data.
     """
-    shelter_id: int
+    shelter_id: str
```
emrameduseljac, line 14: Changing the shelter ID from int to str makes sense, just important to make sure you change the other tests to match what it says here.

emrameduseljac (commented):
Overall, the code is well-organized and it's clear what you're trying to do. The tests are failing, but you did mention it in your PR that it will get resolved later on. I do think adding the filter function was a good idea, to ensure the filtering logic works and all the right arguments are checked as well. Overall, good job, just make sure to make the int to str adjustments so the tests will pass.

PR #237, opened 2025-03-24
Checking for valid shift IDs

### LINKED ISSUE #189: Server side allows to create service commitments for non-existing shifts

emrameduseljac:
Fixes #189 

**What was changed?**

Service commitment creation logic was modified to ensure that only valid shift IDs are used when creating service commitments. Shift validation was added to check if a shift exists in the database before creating a service commitment. Error handling was introduced to return an error message for non-existing shifts.

**Why was it changed?**

Previously, the system was allowing service commitments to be created with invalid shift IDs, so commitments could be made without proper validation of the shift ID. This change ensures that only valid shifts are used for commitments.

**How was it changed?**

The `add_service_commitments` function in `use_cases/add_service_commitments.py` was updated to validate each `service_shift_id` against the existing shifts in the database. It skips the creation of service commitments with invalid shift IDs and returns the proper error message. Only valid commitments are inserted into the database after checking for time overlaps.


server/use_cases/add_service_commitments.py
```diff
@@ -19,22 +19,49 @@ def add_service_commitments(commitments_repo, shifts_repo, commitments):
...
+                "message": f"cannot commit to non-existing shift {shift_id}"
+            })
+            continue
+        valid_commitments.append(commitment)
+        results.append(None) # placeholder to fill
 
-    commitments_as_dict = [c.to_dict() for c in commitments]
+    if not valid_commitments:
+        return [r for r in results if r is not None]
+
```
amalrzv, line 41: Remove this space to get rid of the whitespace and possibly the space after the ]

server/use_cases/add_service_commitments.py
```diff
@@ -19,22 +19,49 @@ def add_service_commitments(commitments_repo, shifts_repo, commitments):
...
+            continue
+        valid_commitments.append(commitment)
+        results.append(None) # placeholder to fill
 
-    commitments_as_dict = [c.to_dict() for c in commitments]
+    if not valid_commitments:
+        return [r for r in results if r is not None]
+        
+    # check for time overlap in valid shift only
+    valid_shifts = [existing_shifts[c.service_shift_id] for c in valid_commitments]
```
amalrzv, line 43: tab this link before "for" to fix the linting message

server/use_cases/add_service_commitments.py
```diff
@@ -19,22 +19,49 @@ def add_service_commitments(commitments_repo, shifts_repo, commitments):
...
+    if check_time_overlap(valid_shifts):
+        return [{
+            "service_commitment_id": None,
+            "success": False
+        }]
+        
+    # insert valid commitments
```
amalrzv, line 50: remove empty line and/or space after ]

server/use_cases/add_service_commitments.py
```diff
@@ -19,22 +19,49 @@ def add_service_commitments(commitments_repo, shifts_repo, commitments):
...
-    if has_overlap:
-        return [{
-            "service_commitment_id": None
-            , "success": False}]
+    valid_commitments = []
+    results = []
+
+    for commitment in commitments:
+        shift_id = commitment.service_shift_id
+        if shift_id not in existing_shifts:
```
amalrzv, line 29: maybe check if existing shifts is being populated how you want it to be to help your troubleshooting

server/use_cases/add_service_commitments.py
```diff
@@ -46,3 +75,99 @@ def check_time_overlap(shifts):
...
+        assert result[1]["success"] is False
+        assert result[1]["message"] == (
+            "cannot commit to non-existing shift this-id-does-not-exist"
+        )
+        print(f"Mixed shifts -> {result}\n")
+
+    # run all tests
+    test_valid_shift_commitment()
+    test_invalid_shift_commitment()
+    test_mixed_shifts_commitment()
```
kate-holdener, line 173: @emrameduseljac, please move the test code to `server/tests/use_cases/add_service_commitments.py`.

server/use_cases/add_service_commitments.py
```diff
@@ -46,3 +75,99 @@ def check_time_overlap(shifts):
...
+            "mocked-valid-shift-id"
+        )
+        shifts_repo.get_shifts.return_value = [valid_shift]
+
+        valid_commitment = Mock()
+        valid_commitment.service_shift_id = "mocked-valid-shift-id"
+
+        valid_commitment.to_dict.return_value = (
+            {"_id": "commitment-id"}
+        )
```
kate-holdener, line 99: Commitment repo assigns the commitment-id when a commitment is added to the repo. When we send a commitment to the add_service_commitments use case, the commitments do not have an id. You'll want to modify your code such that commitments_repo returns ['commitment-id'] when insert_service_commitments is called.

server/use_cases/add_service_commitments.py
```diff
@@ -46,3 +75,99 @@ def check_time_overlap(shifts):
...
+        shifts_repo = Mock()
+        commitments_repo = Mock()
+
+        shifts_repo.get_shifts.return_value = []
+
+        invalid_commitment = Mock()
+        invalid_commitment.service_shift_id = "this-id-does-not-exist"
+
+        invalid_commitment.to_dict.return_value = (
+            {"_id": "commitment-id"}
```
kate-holdener, line 120: When we pass a service commitment to add_service_commitments use case, the commitment does not have an id.

server/use_cases/add_service_commitments.py
```diff
@@ -46,3 +75,99 @@ def check_time_overlap(shifts):
...
+        )
+        result = add_service_commitments(commitments_repo, shifts_repo, (
+            [invalid_commitment])
+        )
+
+        assert result[0]["success"] is False
+        assert result[0]["message"] == (
+            "cannot commit to non-existing shift this-id-does-not-exist"
+        )
+        print(f"Invalid shift ID -> {result}\n")
```
kate-holdener, line 130: You should also add an assert verifying that `insert_service_commitments` was NOT called on the commitments_repo

server/use_cases/add_service_commitments.py
```diff
@@ -46,3 +75,99 @@ def check_time_overlap(shifts):
...
+        invalid_commitment = Mock()
+        invalid_commitment.service_shift_id = (
+            "this-id-does-not-exist"
+        )
+        valid_commitment.to_dict.return_value = (
+            {"_id": "valid-commitment-id"}
+        )
+        invalid_commitment.to_dict.return_value = (
+            {"_id": "invalid-commitment-id"}
+        )
```
kate-holdener, line 156: When we send service commitments to the add_service_commitment use case, they do not have an _id. The _id is added later. You'll want to have your commitments_repo return the id values from the call to insert_service_commitments.

kate-holdener (changes requested):
Nice job adding test cases. The test cases need to be moved to the server/tests directory (I noted the exact file in my inline comments). There are small changes needed in the test cases.

server/tests/use_cases/test_add_service_commitments.py
```diff
@@ -0,0 +1,106 @@
+"""
+test_add_service_commitments.py: Tests for the add_service_commitments use case.
+"""
+
+from use_cases.add_service_commitments import add_service_commitments
+
+if __name__ == "__main__":
+    from unittest.mock import Mock
+
+    # valid shift ID (successful case)
```
kate-holdener, line 10: @emrameduseljac your test cases do not run as part of our test suite because you have them hiding behind 
```
if __name__=="__main__"
```
We want these test cases to run along side other test. To run all test cases locally on your computer, in the server directory, set PYTHONPATH:
`export PYTHONPATH=$(pwd)`
Then run test cases with `pytest`
You will see a listing of all files that ran. When I run pytest on my computer, your test cases do not run. 
Remove the "main" and just include the test case functions in this file. Run the tests to verify that they pass.

server/use_cases/add_service_commitments.py
kate-holdener, line 1: There is a bug in this code somewhere. 
When I try to create a service commitment for an existing shift, I get an error that I cannot commit to a non-existing shift.

Here is a command I use to create a valid shift:
```
--data '[{"shelter_id":22222, "shift_start": 100000, "shift_end": 200000}]'
```
The response from this command I get is:
```
{
    "success": true,
    "service_shift_ids": [
        "67e93f34659f2301057e136d"
    ]
}
```
Then when I try to commit to this shift with:
```
--data '[{"service_shift_id": "67e93f34659f2301057e136d"}]'
```
I get this response:
```
[
    {
        "service_commitment_id": null,
        "success": false,
        "message": "cannot commit to non-existing shift 67e93f34659f2301057e136d"
    }
]
```

kate-holdener (changes requested):
@emrameduseljac Your tests do not run as part of our test suite. Additionally, the logic for creating service commitments does not allow me to commit to **existing** shifts.

kate-holdener (changes requested):
There is a new problem, which looks like a regression from main. When I try to create two commitments for the same valid shift:
```
--data '[{"service_shift_id": "67e93f34659f2301057e136d"},{"service_shift_id": "67e93f34659f2301057e136d"}]'
```
I get only one element in the response array:
```
[
    {
        "service_commitment_id": null,
        "success": false
    }
]
```

When I do the same thing in the main branch, I get two elements.
Technically, I shouldn't be able to create two commitments for the same shift, but let's not try to fix this bug here. The bug is that I only get one value for two attempted commitments, while I should be getting two values.

server/domains/service_shift.py
```diff
@@ -33,7 +32,10 @@ def from_dict(cls, d):
         """
         Creates an instance of ServiceShift from a dictionary.
         """
-        return cls(**d)
+        id_ = str(d.pop("_id", None)) if "_id" in d else None
+        obj = cls(**d)
+        obj.set_id(id_)
```
kate-holdener, line 37: @emrameduseljac what's this for?

server/repository/mongo/service_shifts.py
```diff
@@ -87,5 +87,6 @@ def get_shifts(self, shift_ids):
         Returns:
             list: A list of ServiceShift objects.
         """
-        shifts = self.collection.find({"_id": {"$in": shift_ids}})
+        object_ids = [ObjectId(sid) for sid in shift_ids]
+        shifts = list(self.collection.find({"_id": {"$in": object_ids}}))
```
kate-holdener, line 91: @emrameduseljac These changes probably have to do with the changes you did in domains/service_shift.py. Can you explain why they are needed.

The code generally works and tests pass, I'm just not understanding this detail.

PR #238, opened 2025-03-24
issue 233

### LINKED ISSUE #233: API must require proper authentication and authorization
Our API endpoints sometime check proper authentication (based on the authorization token passed with the request). However, not all API endpoints do this. 

Additionally, once the user is authenticated, the shelter and system admin API endpoints need to check if the user is authorized to make the request. This can now be done using the `is_authorized` function of `server/use_cases/authorization/is_authorized.py`

Review all endpoints the following files and document (in the code) appropriate permissions needed:
- server/application/rest/shelter.py
- server/application/rest/service_shifts.py

Update these endpoints to include proper authentication and authorization. If the user is not authorized to make the request, return a response status associated with ResponseTypes.UNAUTHORIZED.

strejoarciles:
Fixes #233 

**What was changed?**

Authentication and authorization checks were added to the shelter.py and service_shifts.py route handlers within the server/application/rest/ directory. A helper function was also added to extract the user's email from the Authorization header in API requests. This ensures only authenticated and authorized users can access or modify shelter and service shift data.

*Here, describe what part of the application you changed (e.g. login page, database, etc.). If possible, mention what specific components were added, removed, or modified.*
A utility function get_user_email_from_auth_header() was created and stored in a new file:
server/application/rest/auth_helpers.py
This function extracts the user's email from a Bearer token in the Authorization header.

Updated Files:
server/application/rest/shelter.py
server/application/rest/service_shifts.py

**Why was it changed?**
Previously, many API endpoints lacked proper authentication and authorization enforcement. This created a security vulnerability, allowing potentially unauthorized users to access or manipulate sensitive data. By introducing authorization checks using the is_authorized function, the system now validates whether a user has permission to access or manage specific resources (e.g., a shelter or system-level data). This change helps maintain data integrity and restricts functionality based on user roles and access rights.


Taktar (commented):
There are several long lines violating the 80-character limit (C0301). These can be split into multiple lines for readability and compliance

Taktar:
Missing import statements caused undefined-variable (E0602) errors — for example:

json, request, Response, ServiceShiftJsonEncoder, shift_add_use_case, etc. are used but not imported.

Be sure to re-add these imports in service_shifts.py and shelter.py.

Taktar:
Module Missing (Test Errors)
The module repository.mongo.user is imported but does not exist.
Maybe either create this module or update the import path if the repo is in a different location.

This causes ModuleNotFoundError in all related test files (test_shelter.py, test_service_shift.py, test_work_shift.py).

kate-holdener (changes requested):
@strejoarciles I think we will want to distinguish the responses between authentication failure and authorization failure. 

1. If we cannot get a valid user from token, let's continue sending UNAUTHORIZED.
2. However, if a user is verified but does not have permissions to access what they are trying to access, let's add a new response code FORBIDDEN. 

UNAUTHORIZED should be mapped to code 401 and FORBIDDEN should be 403. Those changes would be in the `server/application/rest/status_codes.py` file.

server/application/rest/service_shifts.py
```diff
@@ -5,14 +5,25 @@
...
 from flask import Blueprint, request, Response
 from flask_cors import cross_origin
-from use_cases.add_service_shifts import shift_add_use_case
-from use_cases.list_service_shifts_use_case import service_shifts_list_use_case
-from repository.mongo.service_shifts import ServiceShiftsMongoRepo
-from domains.service_shift import ServiceShift
+
+# Application-specific imports grouped by package
 from application.rest.status_codes import HTTP_STATUS_CODES_MAPPING
+from application.rest.auth_utils import get_user_email_from_token
```
kate-holdener, line 11: `get_user_from_token` is defined in `server/authentication/authenticate_user.py`. You'll want to import that function in each file and use it. Note that it's `get_user_from_token`, not `get_user_email_from_token`

server/application/rest/auth_utils.py
kate-holdener, line 1: This file is unnecessary - it adds complexity. Instead of importing this file, other modules should just import get_user_from_token directly from authentication.authenticate_user.

server/application/rest/service_shifts.py
```diff
@@ -66,6 +113,27 @@ def handle_service_shift():
...
+                    "the same shelter"}),
+                    mimetype="application/json",
+                    status=HTTP_STATUS_CODES_MAPPING[
+                        ResponseTypes.PARAMETER_ERROR],
+                )
+            # Get the shelter_id from the first shift
+            shelter_id = shifts_obj[0].shelter_id
+            # Authorization check for adding service
+            # shifts to a specific shelter
+            if not is_authorized(repo, user_email, Resources.SHIFT, shelter_id):
```
kate-holdener, line 130:
```suggestion
            if not is_authorized(repo, user_email, Resources.SHELTER, shelter_id):
```

server/application/rest/service_shifts.py
```diff
@@ -40,6 +78,15 @@ def handle_service_shift():
...
 
+        # Authorization check for viewing service shifts for a specific shelter
+        if not is_authorized(repo, user_email, Resources.SHIFT, shelter_id):
+            return Response(
+                json.dumps({"message": "Unauthorized to view service "
+                "shifts for this shelter"}),
+                mimetype="application/json",
+                status=HTTP_STATUS_CODES_MAPPING[ResponseTypes.FORBIDDEN]
+            )
+
```
kate-holdener, line 89:
```suggestion
```
Anyone who is authenticated should be allowed to view all defined shifts. Volunteer dashboard will use the API endpoint to show volunteers which shifts can be signed up for.

server/application/rest/shelter.py
```diff
@@ -3,31 +3,67 @@
...
+            status=HTTP_STATUS_CODES_MAPPING[
+                ResponseTypes.UNAUTHORIZED])
+    # Get the user email from the token
+    user_email = get_user_from_token(auth_token)
+    if not user_email:
+        return Response(
+            json.dumps({"message": "Invalid authentication token"}),
+            mimetype="application/json",
+            status=HTTP_STATUS_CODES_MAPPING[ResponseTypes.UNAUTHORIZED]
+        )
```
kate-holdener, line 57: Since this block of code is used repeatedly in our API endpoint handling, it would be good to put it into its own function and reuse the function. For example, you an put it in `server/application/rest/auth_utils.py` and call that function from all the places where this code is repeated.

server/application/rest/shelter.py
```diff
@@ -3,31 +3,67 @@
...
 
     if request.method == "GET":
-        # process the GET request parameters
+        # Authorization check for viewing shelters
+        if not is_authorized(repo, user_email, Resources.SHELTER):
+            return Response(
+                json.dumps({"message": "Unauthorized to view shelters"}),
+                mimetype="application/json",
+                status=HTTP_STATUS_CODES_MAPPING[
+                    ResponseTypes.FORBIDDEN])
```
kate-holdener, line 66:
```suggestion
```
We'll need to allow volunteers to view shelters, otherwise they will not know where to go.

server/application/rest/shelter.py
```diff
@@ -38,8 +74,14 @@ def shelter():
             status=HTTP_STATUS_CODES_MAPPING[ResponseTypes.SUCCESS]
         )
     elif request.method == "POST":
+        # Authorization check for adding shelters
+        if not is_authorized(repo, user_email, Resources.SHELTER):
```
kate-holdener, line 78:
```suggestion
        if not is_authorized(repo, user_email, Resources.SYSTEM):
```
Only system admins are allowed to add shelters.

kate-holdener (changes requested):
@strejoarciles I added comments clarifying permissions of various API endpoints. Also, I added some comments for cleaning up the code and removing code duplication.
Looks like our server side tests cases are failing in your branch, please check on those after making the changes.

kungfuchicken:
@kate-holdener Is this change still needed? @strejoarciles are you still working this?

PR #248, opened 2025-04-07
Issue 191

### LINKED ISSUE #191: Server side allows to create overlapping service commitments
**Describe the bug**
Our POST /service_commitment api endpoint allows users to create service commitments with overlapping times. This is undesirable, because we want to prevent users from singing up for service shifts that have overlapping times.

**To Reproduce**
There are two ways to create overlapping commitments.
Option 1 - send overlapping commitments in the same POST request
1. Create two service sifts for the same time slots but for different shelters
```
--data '[
    {"shelter_id":1111, "shift_start": 10, "shift_end": 20}, 
    {"shelter_id":2222, "shift_start": 10, "shift_end": 20}
]'
```
2. Note the service shift ids that got returned. Suppose they are ID1 and ID2
3. Create two commitments for the same user, one for shift with ID1 and another for shift with ID2
```
--data '[{"service_shift_id": ID1}, {"service_shift_id": ID2}]'
```
5. Both commitments will be successfully created

Option 2 - create commitments one at a time
1. Create two service sifts for the same time slots but for different shelters
```
--data '[
    {"shelter_id":1111, "shift_start": 10, "shift_end": 20}, 
    {"shelter_id":2222, "shift_start": 10, "shift_end": 20}
]'
```
2. Note the service shift ids that got returned. Suppose they are ID1 and ID2
3. Create a commitment for service shift with ID1
```
--data '[{"service_shift_id": ID1}]'
```
4. Create a commitment for service shift with ID2
```
--data '[{"service_shift_id": ID2}]'
```
5. Both commitments will be successfully created

**Expected Behavior**
In both cases (Option 1 and Option 2), the second commitment should be rejected with a message 'Overlapping commitment'. The first commitment should be created.

**Suggested approach**
Issue #189, when completed, will ensure that service commitments will only be created for existing service_shift_id. To do this, the `add_service_commitments` function of use_cases/add_service_commitments.py will be modified: commitment_repo and shift_repo will be passed to the function as arguments. 

After issue #189 is resolved, we can improve the `add_service_commitments` function by checking the time slots associated with each commitment. Time slots are defined in the service shifts, and each commitment is associated with a specific shift id. Therefore, we can retrieve all service shifts (from the shift_repo) for the given list of commitments, check their time slots and see if any of them overlap. If they overlap, we reject the overlapping commitment that appears later in the list with a message 'overlapping commitments'.

Additionally, we need to make sure that the new commitments for this user do not overlap with any commitments that already exist in the database. Therefore, we also need to retrieve a list of all commitments the user has and check for overlap with them.

**Success criteria**

- [x] In the scenario described under Option 1 above, the first commitment should get created while the second one should not
- [x] In the scenario described under Option 2 above, the first commitment should get created while the second one should not.

strejoarciles:
Fixes #191 

I changed the service commitment functionality within the API endpoint for creating service commitments. Specifically, I modified the logic in the add_service_commitments function to prevent users from creating overlapping service commitments.

The change was necessary because the original implementation allowed users to sign up for service shifts with overlapping times, which is undesirable. The API was incorrectly accepting multiple commitments for the same time period, either when submitted in a single request or across multiple requests. This fix ensures that a user can only commit to non-overlapping service shifts.

I modified the add_service_commitments function in use_cases/add_service_commitments.py. The key changes include: retrieving the user's existing commitments from the database using fetch_service_commitments, checking for time overlaps between new commitments and existing ones, implementing a one-at-a-time processing approach, and adding robust error handling for test environments. The most important addition was the overlap detection logic that compares shift start and end times to prevent scheduling conflicts.

### DIFF
--- a/server/use_cases/add_service_commitments.py
+++ b/server/use_cases/add_service_commitments.py
@@ -71,17 +71,20 @@ def add_service_commitments(commitments_repo, shifts_repo, commitments):
         current_shift = valid_shifts[i]
         has_overlap = False
         # Check for overlap with existing user shifts
-        shifts_to_check = [current_shift]
-        shifts_to_check.extend(existing_user_shifts)
-        if check_time_overlap(shifts_to_check):
-            has_overlap = True
-        if not has_overlap and final_valid_indexes:
+        for existing_shift in existing_user_shifts:
+            if (current_shift.shift_start < existing_shift.shift_end and
+                current_shift.shift_end > existing_shift.shift_start):
+                has_overlap = True
+                break
+        # Check for overlap with previously approved shifts in this request
+        if not has_overlap:
             approved_shifts = [
                 valid_shifts[j] for j in range(i) if j in final_valid_indexes]
-            shifts_to_check = [current_shift]
-            shifts_to_check.extend(approved_shifts)
-            if check_time_overlap(shifts_to_check):
-                has_overlap = True
+            for approved_shift in approved_shifts:
+                if (current_shift.shift_start < approved_shift.shift_end and
+                    current_shift.shift_end > approved_shift.shift_start):
+                    has_overlap = True
+                    break
         if has_overlap:
             results[idx] = {
                 "service_commitment_id": None,
@@ -91,14 +94,10 @@ def add_service_commitments(commitments_repo, shifts_repo, commitments):
         else:
             final_valid_commitments.append(valid_commitments[i])
             final_valid_indexes.append(idx)
+            # Add current shift to existing_user_shifts
+            # to check against subsequent commitments
+            existing_user_shifts.append(current_shift)
     if not final_valid_commitments:
-        for i in valid_indexes:
-            if results[i] is None:
-                results[i] = {
-                    "service_commitment_id": None,
-                    "success": False,
-                    "message": "Overlapping commitment"
-                }
         return results
     commitments_as_dict = [c.to_dict() for c in final_valid_commitments]
     inserted_commitments = commitments_repo.insert_service_commitments(
--- a/server/use_cases/add_service_commitments.py
+++ b/server/use_cases/add_service_commitments.py
@@ -44,62 +44,44 @@ def add_service_commitments(commitments_repo, shifts_repo, commitments):
             results.append(None)
     if not valid_commitments:
         return results
-    if is_mock_object(existing_shifts) or is_mock_object(shifts_repo):
-        commitments_as_dict = [c.to_dict() for c in valid_commitments]
-        inserted_commitments = commitments_repo.insert_service_commitments(
-            commitments_as_dict)
-        for idx, i in enumerate(valid_indexes):
-            if idx < len(inserted_commitments):
-                results[i] = {
-                    "service_commitment_id": str(inserted_commitments[idx]),
-                    "success": True
-                }
-            else:
-                results[i] = {
-                    "service_commitment_id": None,
-                    "success": False,
-                    "message": "Failed to insert commitment"
-                }
-        return results
+    # Get user ID from first commitment
     user_id = valid_commitments[0].volunteer_id
-    existing_user_commitments = commitments_repo.fetch_service_commitments(
-        user_id=user_id)
-    if existing_user_commitments is None or not hasattr(
-        existing_user_commitments, "__iter__"):
+    # Fetch user's existing commitments
+    try:
+        existing_user_commitments = commitments_repo.fetch_service_commitments(
+            user_id=user_id)
+        if existing_user_commitments is None:
+            existing_user_commitments = []
+    except (AttributeError, TypeError):
         existing_user_commitments = []
     existing_commitment_shift_ids = [str(c.service_shift_id)
-                                     for c in existing_user_commitments
-                                     if c.service_shift_id is not None]
-    if not existing_commitment_shift_ids:
-        existing_user_shifts = []
-    else:
+                                    for c in existing_user_commitments
+                                    if c.service_shift_id is not None]
+    # Get existing shifts
+    existing_user_shifts = []
+    if existing_commitment_shift_ids:
         existing_user_shifts = shifts_repo.get_shifts(
             existing_commitment_shift_ids)
-    valid_shifts = [existing_shifts[c.service_shift_id]
-                    for c in valid_commitments]
+    # Get valid shifts for current commitments
+    valid_shifts = [
+        existing_shifts[c.service_shift_id] for c in valid_commitments]
     final_valid_commitments = []
     final_valid_indexes = []
     for i, idx in enumerate(valid_indexes):
         current_shift = valid_shifts[i]
         has_overlap = False
-        for existing_shift in existing_user_shifts:
-            if has_valid_time_attributes(
-                current_shift) and has_valid_time_attributes(
-                existing_shift):
-                if (current_shift.shift_start < existing_shift.shift_end and
-                    current_shift.shift_end > existing_shift.shift_start):
-                    has_overlap = True
-                    break
-        approved_shifts = [
-            valid_shifts[j] for j in range(i) if j in final_valid_indexes]
-        for approved_shift in approved_shifts:
-            if has_valid_time_attributes(
-                current_shift) and has_valid_time_attributes(
-                approved_shift):
-                if (current_shift.shift_start < approved_shift.shift_end and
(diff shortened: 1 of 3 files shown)


server/use_cases/add_service_commitments.py
```diff
@@ -39,56 +41,157 @@ def add_service_commitments(commitments_repo, shifts_repo, commitments):
...
-                shifts[i].shift_end > shifts[j].shift_start):
-                return True
+            if has_valid_time_attributes(
+                shifts[i]) and has_valid_time_attributes(
+                shifts[j]):
+                if (shifts[i].shift_start < shifts[j].shift_end and
+                    shifts[i].shift_end > shifts[j].shift_start):
+                    return True
     return False
+
```
Taktar, line 165: Love the idea of checking validity of attributes to guard against mocks or malformed input. Nice touch! Maybe add unit tests for these new helpers to ensure edge cases (like partially missing fields or weird mocks) are covered.

server/use_cases/add_service_commitments.py
```diff
@@ -39,56 +41,157 @@ def add_service_commitments(commitments_repo, shifts_repo, commitments):
...
             }
         else:
             results[i] = {
                 "service_commitment_id": None,
-                "success": False
+                "success": False,
+                "message": "Failed to insert commitment"
+            }
+    for i in valid_indexes:
+        if results[i] is None:
```
Taktar, line 137: Nice catch updating the fallback None to an explicit response dict with an error message — this is definitely more informative for debugging and frontend feedback!

server/use_cases/add_service_commitments.py
```diff
@@ -39,56 +41,157 @@ def add_service_commitments(commitments_repo, shifts_repo, commitments):
...
+            results[i] = {
+                "service_commitment_id": None,
+                "success": False,
+                "message": "Overlapping commitment"
             }
-
     return results
 
-
 def check_time_overlap(shifts):
```
kate-holdener, line 145: You have a check_time_overlap function, but this function does not get used. You should use it.

server/use_cases/add_service_commitments.py
```diff
@@ -39,56 +41,157 @@ def add_service_commitments(commitments_repo, shifts_repo, commitments):
...
+        approved_shifts = [
+            valid_shifts[j] for j in range(i) if j in final_valid_indexes]
+        for approved_shift in approved_shifts:
+            if has_valid_time_attributes(
+                current_shift) and has_valid_time_attributes(
+                approved_shift):
+                if (current_shift.shift_start < approved_shift.shift_end and
+                    current_shift.shift_end > approved_shift.shift_start):
+                    has_overlap = True
+                    break
```
kate-holdener, line 102: Instead of redefining the check for time overlap logic here, you should use the check_time_overlap function  you have in this file.

kate-holdener (changes requested):
@strejoarciles I encourage you to test your code changes. When I send a POST request to create a service commitment for a valid shift ID, I get a 400 response status code with a message:
```
{
    "error": "Invalid shift ID: '134134' is not a valid ObjectId, it must be a 12-byte input or a 24-character hex string"
}
```
I added other inline comments for changes that need to be made.

server/use_cases/add_service_commitments.py
```diff
@@ -39,52 +41,99 @@ def add_service_commitments(commitments_repo, shifts_repo, commitments):
...
+        shifts_to_check.extend(existing_user_shifts)
+        if check_time_overlap(shifts_to_check):
+            has_overlap = True
+        if not has_overlap and final_valid_indexes:
+            approved_shifts = [
+                valid_shifts[j] for j in range(i) if j in final_valid_indexes]
+            shifts_to_check = [current_shift]
+            shifts_to_check.extend(approved_shifts)
+            if check_time_overlap(shifts_to_check):
+                has_overlap = True
```
kate-holdener, line 84:
```suggestion
```
It's redundant to do the check for overlap again. I understand why you are doing it - you want to make sure that it doesn't overlap with any of the commitments sent in the same request that you already approved. There is a simpler way to do this. See my suggestion later in this function.

server/use_cases/add_service_commitments.py
```diff
@@ -39,52 +41,99 @@ def add_service_commitments(commitments_repo, shifts_repo, commitments):
...
+            final_valid_commitments.append(valid_commitments[i])
+            final_valid_indexes.append(idx)
+    if not final_valid_commitments:
+        for i in valid_indexes:
+            if results[i] is None:
+                results[i] = {
+                    "service_commitment_id": None,
+                    "success": False,
+                    "message": "Overlapping commitment"
+                }
```
kate-holdener, line 101:
```suggestion
```
I'm pretty sure that this is redundant. Since earlier, when you looped through valid_indexes, if the commitment was overlapping, you already inserted the response for the overlapping commitment into results[i] (on line 86)

server/use_cases/add_service_commitments.py
```diff
@@ -39,52 +41,99 @@ def add_service_commitments(commitments_repo, shifts_repo, commitments):
...
-                "success": False
+                "success": False,
+                "message": "Failed to insert commitment"
+            }
+    for i in valid_indexes:
+        if results[i] is None:
+            results[i] = {
+                "service_commitment_id": None,
+                "success": False,
+                "message": "Overlapping commitment"
```
kate-holdener, line 123: This also seems redundant, you already marked the commitments as overlapping, when you discovered that they are overlapping (on line 86)
```suggestion
```

kate-holdener (changes requested):
@strejoarciles 
Nice job on the progress. The code works for the cases specified in the issue.
There is a lot of redundant code that makes it difficult to read. I added my comments with proposed changes on how to simplify it.

Additionally, there is a bug: we are only considering overlaps if they are for the same shelter. We should not allow to create commitments for the same volunteer that overlap, regardless of whether they are in the same shelter or not (the same person cannot be volunteering in two different places at the same time).

server/use_cases/add_service_shifts.py
```diff
@@ -39,11 +39,15 @@ def shift_add_use_case(repo, new_shifts):
...
         for other_shift in shifts[i + 1:]:
-            if (max(shift.shift_start, other_shift.shift_start) <
-                min(shift.shift_end, other_shift.shift_end)) and \
-               (shift.shelter_id == other_shift.shelter_id):
+            # Check time overlap
+            time_overlap = (
+                max(shift.shift_start, other_shift.shift_start) <
+                    min(shift.shift_end, other_shift.shift_end))
+            # Only consider it an overlap if at the same shelter
+            if time_overlap and shift.shelter_id == other_shift.shelter_id:
```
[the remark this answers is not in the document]
kate-holdener: Oh, nevermind, I thought I was reviewing service_commitments code, but now I see that it's a change to add_service_shifts.py

server/use_cases/add_service_commitments.py
```diff
@@ -39,52 +41,98 @@ def add_service_commitments(commitments_repo, shifts_repo, commitments):
...
+    final_valid_indexes = []
+    for i, idx in enumerate(valid_indexes):
+        current_shift = valid_shifts[i]
+        has_overlap = False
+        # Check for overlap with existing user shifts
+        for existing_shift in existing_user_shifts:
+            if (current_shift.shift_start < existing_shift.shift_end and
+                current_shift.shift_end > existing_shift.shift_start):
+                has_overlap = True
+                break
```
kate-holdener, line 78: What happened to calling check_time_overlap? You used to call it here, but now you duplicated the code that's in check_time_overlap in this function.

kate-holdener (changes requested):
@strejoarciles You didn't address the code duplication problems I pointed out in my last review. Also, you added some more code duplication in this version, instead of reusing check_time_overlap function. Please address all the concerns I pointed out.

server/use_cases/add_service_commitments.py
```diff
@@ -39,52 +41,98 @@ def add_service_commitments(commitments_repo, shifts_repo, commitments):
...
+            results[i] = {
+                "service_commitment_id": None,
+                "success": False,
+                "message": "Overlapping commitment"
             }
-
     return results
 
-
 def check_time_overlap(shifts):
```
kate-holdener, line 126: You should change this function to be:
```suggestion
def check_time_overlap(shift_to_check, other_shifts):
```
Have this function return true if shift_to_check overlaps with any of the shifts in other_sifts.
Then use this function in your code

PR #262, opened 2025-04-25
Fixing Issue 252

### LINKED ISSUE #252: Support GET /schedule API endpoint on the server

To allow shelter admins to use their pre-loaded shift schedules when opening shelters, we'll need to retrieve those schedules through the server. To do this, we'll add a GET /schedule?shelter_id=<SHELTER_ID> API endpoint. This endpoint will return the service shift objects found in the `schedules` collection of our database. As with issue #251, we can probably reuse service_shift use cases for handling this as well as ServiceShiftsMongoRepo database (just need to point that database to the schedules collection).

emrameduseljac:
Fixes #252

**What was changed?**

* Added a new API endpoint `/schedule` to allow shelter admins to retrieve pre-loaded shift schedules.
* Created a new repository `ScheduleMongoRepo` to interact with the `schedules` collection in the database.
* Added a new use case `get_schedule_shifts_use_case.py` to handle the retrieval of schedule shifts based on shelter ID.
* Modified `server/application/rest/service_shifts.py` to register the new `/schedule` route.

**Why was it changed?**

Previously, shelter admins could not load their pre-defined schedule shifts when opening shelters. This change adds the necessary backend support to fetch pre-saved schedules associated with a given shelter ID.

**How was it changed?**

* New file: `server/use_cases/get_schedule_shifts_use_case.py`, added a function retrieve shifts by shelter ID.
* New file: `server/repository/mongo/schedule_repo.py`, created a new MongoDB repository class to point to the `schedules` collection.
* Modified file: `server/application/rest/service_shifts.py`, imported the new repository and use case. Added a new route `/schedule` that accepts a GET request with a `shelter_id` parameter.

### DIFF
--- a/server/application/rest/service_shifts.py
+++ b/server/application/rest/service_shifts.py
@@ -12,6 +12,8 @@ from domains.service_shift import ServiceShift
 from application.rest.status_codes import HTTP_STATUS_CODES_MAPPING
 from responses import ResponseTypes
 from serializers.service_shift import ServiceShiftJsonEncoder
+from use_cases.get_schedule_shifts_use_case import get_schedule_shifts_use_case
+from repository.mongo.schedule_repo import ScheduleMongoRepo
 
 service_shift_bp = Blueprint("service_shift", __name__)
 
@@ -87,3 +89,23 @@ def handle_service_shift():
             mimetype="application/json",
             status=status_code,
         )
+
+@service_shift_bp.route("/schedule", methods=["GET"])
+@cross_origin()
+def handle_schedule_shift():
+    shelter_id = request.args.get("shelter_id")
+    if not shelter_id:
+        return Response(
+            json.dumps({"error": "shelter_id is required"}),
+            mimetype="application/json",
+            status=400
+        )
+        
+    repo = ScheduleMongoRepo()
+    shifts = get_schedule_shifts_use_case(repo, shelter_id)
+
+    return Response(
+        json.dumps(shifts, cls=ServiceShiftJsonEncoder),
+        mimetype="application/json",
+        status=200
+    )
--- a/server/repository/mongo/schedule_repo.py
+++ b/server/repository/mongo/schedule_repo.py
@@ -1,0 +1,11 @@
+from domains.service_shift import ServiceShift
+from config.mongodb_config import get_db
+
+class ScheduleMongoRepo:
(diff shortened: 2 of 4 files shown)


server/application/rest/service_shifts.py
```diff
@@ -87,3 +89,23 @@ def handle_service_shift():
...
+        )
+        
+    repo = ScheduleMongoRepo()
+    shifts = get_schedule_shifts_use_case(repo, shelter_id)
+
+    return Response(
+        json.dumps(shifts, cls=ServiceShiftJsonEncoder),
+        mimetype="application/json",
+        status=200
+    )
```
kate-holdener, line 111: I think this should be in a separate file: schedule.py

server/repository/mongo/schedule_repo.py
```diff
@@ -0,0 +1,11 @@
...
+from config.mongodb_config import get_db
+
+class ScheduleMongoRepo:
+    def __init__(self):
+        self.db = get_db()
+        self.collection = self.db.schedules
+
+    def list_by_shelter_id(self, shelter_id):
+        shifts = self.collection.find({"shelter_id": shelter_id})
+        return [ServiceShift.from_dict(s) for s in shifts]
```
kate-holdener, line 11: I wonder if ScheduleMongoRepo could extend ServiceShiftsMongoRepo. Essentially the functionality is the same and it will be working with the same data structure: ServiceShift. The only thing that you'll need to redefine is the constructor (which you are already doing). Then, in the /schedule API endpoint, you'd create a ScheduleMongoRepo instance and pass it to the service_shifts_list_use_case. You won't need to have a separate use case for the schedule, you can just reulse the service_shift_list_use_case.

kate-holdener (changes requested):
I added a few inline comments: Since /schedule API endpoint will be working with ServiceShift data structure, we can reuse the existing use cases, such as service_sift_list_use_case. We'll just need to pass it a ScheduleMongoRepo, instead of ServiceShiftRepo. If ScheduleMongoRepo extends ServiceShiftRepo, this should work.

Also, this code cannot be tested manually right now because we don't have any shifts in the schedule. The only way I see to test it is to add test cases where we have a mock use case implementation that returns ServiceShift objects from the service_shift_list_use_case.

PR #263, opened 2025-04-27
schedule api endpoint

### LINKED ISSUE #254: Update /set-shifts to use data from the /schedule API endpoint

When opening a shelter, right now hard-coded schedule data is used. Instead, we want to retrieve the shelter-specific schedule from the server (from the GET /schedule?shelter_id=<SHELTER_ID> API endpoint), and use that data to open shelters.

strejoarciles:
Fixes #254 

The /set-shifts feature was updated to use schedule data fetched from the server's /schedule API endpoint instead of using hard-coded schedule data. 

Previously, the Schedule component used hard-coded schedule data from the ScheduleData.js file, which limited flexibility and didn't allow for shelter-specific shift templates. This change enables the system to retrieve shelter-specific schedule data from the server, providing a more dynamic and customizable experience for shelter administrators.

Added a new getSchedule function to src/api/shelter.js to fetch schedule data from the /schedule?shelter_id=<SHELTER_ID> API endpoint
Modified src/components/shelter/Schedule.jsx to:
Remove the import of the hard-coded ScheduleData file
Add useEffect hook to fetch schedule data when the component mounts
Add loading and error states to handle API request lifecycle
Transform the API response to match the expected data structure
Use the dynamically fetched data throughout the component instead of the hard-coded data

### DIFF
--- a/client_app/src/api/shelter.js
+++ b/client_app/src/api/shelter.js
@@ -15,5 +15,10 @@ export const shelterAPI = {
             body: JSON.stringify(data),
         });
         return response;
+    },
+    // Add this function to the shelterAPI object in shelter.js
+    getSchedule: async (shelterId) => {
+    const response = await fetchClient(`/schedule?shelter_id=${shelterId}`);
+    return response;
     }
 }
--- a/client_app/src/api/shelter.js
+++ b/client_app/src/api/shelter.js
@@ -15,10 +15,38 @@ export const shelterAPI = {
             body: JSON.stringify(data),
         });
         return response;
-    },
-    // Add this function to the shelterAPI object in shelter.js
-    getSchedule: async (shelterId) => {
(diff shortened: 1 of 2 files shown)


kate-holdener (changes requested):
@strejoarciles I left a couple of inline comments. Also, I'm wondering how we can test this (without the backend being ready to give you the data).

Ideally, it would be with a new test suite for the front end. If that's too challenging, then
maybe you can temporarily add a mock implementation of getSchedule (in api/shelter.js) that returns a list of hard-coded service shift objects (in the same format that the backend API will send it).

kate-holdener (changes requested):
I'm not sure how to see your changes in the client side application. Please work with the rest of the team to integrate your changes into the 'schedule' feature of the application.

PR #267, opened 2025-04-28
added post/scheduling api endpoints

### LINKED ISSUE #250: Allow shelter admins to define repeatable schedules
Shelter admins can open shelter through the shelter dashboard. When admin opens a shelter, a pre-defined shift schedule is loaded into the calendar. This pre-defined shift schedule is currently hard-coded. We want to support dynamically defined shift schedules.

In the "Settings" page of the shelter dashboard, the shelter admin should see an option to "Define Repeatable Shifts". In that option, the user should be able to define the shifts for a given "work day". Note that it's possible for a shift to start in the evening of one day and end in the morning of the next day. The timestamp format in this schedule is milliseconds since midnight (on the day when the user is opening the shelter).

Once the user defines what a typical work day would look like, they would submit it to the server for processing. The server will need to validate the data, and save it in the "schedules" collection of the database (note that this will be a new collection).

Once the schedule is saved, when the user goes to open a shelter, that shelter's repeatable schedule should be loaded (not the hard-coded schedule).

amalrzv:
Fixes #250 

Created a new endpoint for handling repeatable schedules. The implementation stores schedule templates in a dedicated collection and reuses our existing service shift code. need to still add tests and implement the schedule generation functionality do actual create the service shifts from the template

### DIFF
--- a/server/application/rest/schedule.py
+++ b/server/application/rest/schedule.py
@@ -25,14 +25,14 @@ def create_schedule():
             #making sure the shift has required fields
             if 'timestamp' not in shift or 'service' not in shift:
                 return jsonify({'error'
-                                : 'Each shift must have timestamp and service fields'}),
+                                : 'Shift requires timestamp + service fields'}),
                 400
             #timestamp here is "milliseconds since midnight" instead of epoch
-            # check this, trying to insert the shift into the schedule collection
+            # check this, trying to insert shift into  schedule collection
             result = schedule_repo.insert(shift)
             inserted_ids.append(str(result))
         #return success with the IDs of created schedules
-        return jsonify({'message': 'Schedule created successfully', 
+        return jsonify({'message': 'Schedule created successfully',
                         'ids': inserted_ids}), 201
     except Exception as e:
         return jsonify({'error': str(e)}), 500
--- a/server/application/rest/schedule.py
+++ b/server/application/rest/schedule.py
@@ -18,19 +18,22 @@ def create_schedule():
         #parsing the request body as JSON
         shifts = request.get_json()
         if not shifts or not isinstance(shifts, list):
-            return jsonify({'error': 'Expected a list of service shifts'}), 400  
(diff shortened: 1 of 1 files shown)


server/application/rest/schedule.py
```diff
@@ -0,0 +1,39 @@
...
+    try:
+        #parsing the request body as JSON
+        shifts = request.get_json()
+        if not shifts or not isinstance(shifts, list):
+            return jsonify({'error': 'Expected a list of service shifts'}), 400
+        #going to process every shift
+        inserted_ids = []
+        for shift in shifts:
+            #making sure the shift has required fields
+            if 'timestamp' not in shift or 'service' not in shift:
```
kate-holdener, line 26: We don't have a `timestamp` field in the shifts. The data that will be sent will match the data of a service_shift structure.

server/application/rest/schedule.py
```diff
@@ -0,0 +1,39 @@
...
+                return jsonify({'error'
+                                : 'Shift requires timestamp + service fields'}),
+                400
+            #timestamp here is "milliseconds since midnight" instead of epoch
+            # check this, trying to insert shift into  schedule collection
+            result = schedule_repo.insert(shift)
```
kate-holdener, line 32: Don't insert into the repo here. Go through the appropriate use case to add the shifts into the repo.

server/application/rest/schedule.py
```diff
@@ -0,0 +1,39 @@
...
+            inserted_ids.append(str(result))
+        #return success with the IDs of created schedules
+        return jsonify({'message': 'Schedule created successfully',
```
kate-holdener, line 35: Please use the same response object as used in other API endpoint, for consistency.

kate-holdener (changes requested):
@amalrzv I recommend that you and @emrameduseljac create an integration branch and merge your work into that branch. This is because you two are working on the POST and GET endpoints for the same data.

PR #286, opened 2025-05-04
Merging issue 251 and issue 252 branches together

### LINKED ISSUE #251: Support POST /schedule API endpoint on the server

To support dynamically defined repeatable schedules, we'll need to implement the POST /schedule API endpoint on the server. Requests to this endpoint are expected to have payload in JSON format - a list of service shift objects to be used for future service shifts. I think it's possible to use the existing infrastructure we have for processing service_shift objects to support this. However, instead of saving these repeatable shifts in our service_shift collection, we should save it in the schedule collection. This way, they'll be stored separately and won't be confused with actual service shifts for which volunteers can sign up.

The timestamp in these shifts will be "milliseconds since midnight" instead of "milliseconds since Epoch", but existing service_shift use cases should be able to handle it - it's just an int from their perspective.

We can even reuse ServiceShiftsMongoRepo, but we'll need to refactor it slightly: instead of a no argument constructor we'll want to pass the name of the collection we want to use to the constructor of that class.

### LINKED ISSUE #252: Support GET /schedule API endpoint on the server

emrameduseljac:
Fixes #251 
Fixes #252 

**What was changed?**

Branches from resolving issues 251 and 252 are now merged. 

**Why was it changed?**

It was changed because we had two separate branches resolving issues regarding POST and GET requests, which could make it confusing and more complicating to resolve without making sure there are no conflicts or overlaps.

**How was it changed?**

Merged branch issue252-getendpoint and RepeatScheduleSupport together.

### DIFF
--- a/server/application/app.py
+++ b/server/application/app.py
@@ -10,6 +10,7 @@ from application.rest.shelter import shelter_blueprint
 from application.rest.service_shifts import service_shift_bp
 from application.rest.authorization.authorization import authorization_blueprint
 from application.rest.login import login_blueprint
+from application.rest.schedule_get import schedule_bp
 
 from config import mongodb_config
 import os
@@ -32,6 +33,7 @@ def create_app(config_name = "development"):
     app.register_blueprint(service_shift_bp)
     app.register_blueprint(authorization_blueprint)
     app.register_blueprint(login_blueprint)
+    app.register_blueprint(schedule_bp)
 
     load_dotenv()  # Load environment variables from the .env file
 
--- a/server/application/rest/schedule_get.py
+++ b/server/application/rest/schedule_get.py
@@ -1,0 +1,30 @@
+import json
+from flask import Blueprint, request, Response
+from flask_cors import cross_origin
+from repository.mongo.schedule_repo import ScheduleMongoRepo
+from use_cases.list_service_shifts_use_case import service_shifts_list_use_case
+from application.rest.status_codes import HTTP_STATUS_CODES_MAPPING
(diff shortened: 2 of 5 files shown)


server/use_cases/schedule_add_use_case.py
kate-holdener, line 1: You should be using add_service_shifts use case. That function has a lot of useful checks, for example checks for overlaps.

kate-holdener (changes requested):
When I try to create a schedule using a POST /schedule API endpoint, I get a 405 response - method not allowed:
```
curl --location '127.0.0.1:5000/schedule' \
--data '[
    {
      "name": "Afternoon Shift",
      "shift_start": 46800000, 
      "shift_end": 61200000,  
      "shelter_id": "6806746a227bca1c924f62eb"
    }
  ]'
```

server/use_cases/schedule_add_use_case.py
amalrzv: Changed this and was able to get it to to be successful; also got this when i tried making a shift at the same time verifying overlapping shifts : {"service_shift_id": null, "success": false, "message": "overlapping shift"}%

PR #292, opened 2025-05-13
Update undo shifts

Taktar:
What was changed?

This update modifies the Schedule.jsx file in the client_app/src/components/shelter/ directory. Specifically, it updates the logic for rendering "Open Shift" events and adds functionality to toggle shifts on and off (i.e., cancel them) by clicking the same event again.

No other files outside of Schedule.jsx were changed. When I pulled from main there were files that looked different on my vscode compared to what I saw on main on github. this happened last time too for some reason, but I knew I pulled the latest version of main properly because everything else matched. And I know these files that looked different wasn't due to me adding or taking away anything because they are files I don't normally work or haven't ever worked on. I just want to clarify this because it will say I made changes to multiple files, but that was just me adding back what should have been there according to the files on main in github.

Why was it changed?

Previously, once a user clicked "Open Shift" on a specific day, all default shifts would be added, but there was no way to undo or cancel those shifts before confirming them. This update adds a "Cancel Shift" toggle so users can remove any default shifts they previously opened on a given day. This keeps the interface more flexible and user-friendly.

How was it changed?

openedDays was refactored from an array to a Set to allow more efficient toggling and lookup.

The handleOpenDate function was updated to support toggling shifts on and off for a given day.

The openShiftEvents logic was changed to dynamically display either “Open Shift” or “Cancel Shift” depending on whether shifts for that day have already been opened.

Only Schedule.jsx was edited — no other files or components were modified.
Screenshot 2025-05-13 at 10 47 42 AM


kate-holdener (changes requested):
The code works now, nice job.

Please revert your changes on AddUserForm.css and UpcomingRequests.tsx, because they are outside of the scope for this issue.

Taktar:
> The code works now, nice job.
> 
> Please revert your changes on AddUserForm.css and UpcomingRequests.tsx, because they are outside of the scope for this issue.

got rid of UpcomingRequests.tsx because it isn't in Main. AddUserForm.css should be good too.

Taktar:
Screenshot 2025-05-13 at 11 17 25 PM

I reverted all the files just like it was instructed in the email and now im getting that error 400 again.

kate-holdener (approved):

kate-holdener:
Nice job, @Taktar !

PR #298, opened 2025-05-31
Issue 297

### LINKED ISSUE #297: Add Google Authentication
Our current authentication system relies on pre-configured user emails. This was done for development purpose and will not work in production environment. We want to uniquely identify and authenticate each user. One way to do this is to use Google Authentication.

Add a "Sing in with Google" to the client side of the application, taking the user to the Google Sing-in page. Use the standard "Sing in with Google" implementation to support this feature. 

**Describe alternatives you've considered**
In the future, we might add other "Sign in with ..." options, but we'll start with Google for now. An alternative is to implement our own authentication, but it will be less secure and will require work that we won't need to do if we use "Sign in with Google".

kate-holdener:
Fixes #297 

**What was changed?**

Added "Sign In with Google" tot he client side home page and the necessary backend details to support Google Authentication. 

**Why was it changed?**

This will allow any user with a Google Account to sign in to our application without creating an account. All users will have access to volunteer dashboard. Users with special privileges will have access to shelter dashboards and/or system admin dashboard. 

**How was it changed?**

On the client side, "Sign In with Google" is implemented using GoogleLogin component of @react-oauth/google. When the sign in is successful, an id_token is returned from Google. This id_token is sent to our server backend's /login API endpoint. This API endpoint verifies the id_token with Google, and extracts user information from it (currently, only email address is used by our backend). The backend then generates a JWT token, based on the user's email address and gives that token an expiration time of 1 day. This token is sent back to the client. On each subsequent request, the client sends this token in the Authorization packet header. We verify this token on the backend (without Google, because this is our token).

The following values need to added to the .env file:
GOOGLE_CLIENT_ID
JWT_SECRET


server/run_dev_server.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 export FLASK_ENV="pre-production"
 export FLASK_CONFIG="development" 
-flask run --debug -h 0.0.0.0 -p 5000
\ No newline at end of file
+flask run --debug -h 0.0.0.0 -p 5001
```
kate-holdener, line 4:
```suggestion
flask run --debug -h 0.0.0.0 -p 5000
```

Issue #313, opened 2025-07-24
"Your Impact" screen on Volunteer Dashboard looks bad


kate-holdener:
**Describe the bug**
When I navigate to "Your Impact" screen of Volunteer Dashboard, here's what I see

<img width="1309" height="686" alt="Image" src="https://github.com/user-attachments/assets/d008d5b4-a76e-4097-8c69-3e628b53653b" />

That looks really bad. The screen needs to be updated to be visually appealing. Additionally, it shouldn't count any shift that has not yet ended as "hours served" or "shelters served" (it currently counts that, as soon as the shift has started).

PR #319, opened 2025-09-15
Issue #313: fix Your Impacts now does not include in-progress shifts,…

Reva335:
Fixes #issue_number 313

**What was changed?**

Frontend (UI): client_app/src/components/volunteer/Impact.js

Exclude in-progress shifts from totals only count shifts with a real shift_end that is not in the future.

More accurate total hours calculation sum fractional hours, display to 1 decimal .

Unique shelters count handles shelter as id or object with _id.

Backend (use-case): server/use_cases/list_service_commitments.py

New completed-only guard: filters commitments to those whose associated shift has shift_end <= now.

**Why was it changed?**
The “Your Impact” screen was over-counting because in-progress shifts were being included in:

Total hours served

Shelters served

This PR ensures the metrics reflect only completed work, which matches user expectations and the product requirement.

**How was it changed?**
Frontend

Added helper onlyCompleted(shift) to filter out ongoing shifts.

Updated calculateTotalHours to sum milliseconds → hours and display with toFixed(1).

Updated calculateUniqueShelters to handle both shelter ids and objects.

client_app/src/components/volunteer/Impact.js

Backend

In list_service_commitments_with_shifts(...), after loading commitments + shifts, added:

now = datetime.utcnow()

completed_shift_ids = { s.get_id() for s in shifts if s.shift_end and s.shift_end <= now }

Filter commitments/shifts to completed_shift_ids before applying time filters.

server/use_cases/list_service_commitments.py


client_app/src/components/shelter/ShiftDetailsTable.js
```diff
@@ -92,7 +92,7 @@ export const ShiftDetailsTable = props => {
             arrow
             followCursor
             enterTouchDelay={0}
-            leaveTouchDelay={5000}>
+            leaveTouchDelay={5001}>
```
lockermanwxlf, line 95: Code review: This may have been unintentionally changed in your grep. It doesn't seem related to the port.

client_app/src/components/volunteer/Impact.js
```diff
@@ -1,54 +1,62 @@
...
-        <h4>Lives Touched</h4>
-        <p>Too many to count</p>
-      </div>
-      <div className="card">
-        <h4>Shelters served</h4>
-        <p>{impactData.sheltersServed}</p>
+    <div style={{ maxWidth: 720, margin: "0 auto", padding: "1rem" }}>
+      <h3 style={{ textAlign: "center", marginBottom: "1rem" }}>Impact Created</h3>
+      <div className="card" style={{ borderRadius: 12, padding: "12px 16px", marginBottom: 12, boxShadow: "0 2px 10px rgba(0,0,0,0.06)" }}>
+        <h4 style={{ margin: 0 }}>Total hours served</h4>
```
lockermanwxlf, line 52: I think the PR is good, and I think the UI looks fine unless the client gives us anything specific to change, but I believe @Supraja050202 mentioned that the Lives Touched should be removed since we aren't measuring it.

lockermanwxlf (changes requested):

Supraja050202 (commented):
Core functionality is correct, metrics are now accurate, and the backend filtering is solid. But for the UI lets make it more effective and I will also try to mention the ideas from the client in The Issue.

client_app/src/components/shelter/ShiftDetailsTable.js
kungfuchicken: good catch

PR #346, opened 2025-10-19
POST on profile user info edit page submission

### LINKED ISSUE #344: Implement POST Request for Volunteer Profile Creation (POST /volunteer/profile)
Title: Implement POST Request for Volunteer Profile Creation (POST /volunteer/profile)
Labels: frontend, api-integration

Description:
Currently, the Volunteer Profile Edit page uses a simulated save to update profile data. This task is to replace that simulation with a real POST request to the backend /volunteer/profile endpoint.

Tasks:

- Create a createUserProfile function in volunteerProfileAPI.js to send a POST request.

- Update handleSave in ProfileSettings.jsx to call createUserProfile instead of using simulated logic.

- Ensure the read-only email field is not included in the POST payload.

- Display loading indicators while the request is pending.

- Show a success message if the POST succeeds or an error message if it fails.

- Update the local profileData state with the response from the backend.

Acceptance Criteria:

- [x] Submitting the Profile Edit form sends a real POST request to /volunteer/profile.

- [x] The volunteer’s profile is created or updated correctly on the backend.

- [x] The read-only email field is never sent in the request.

- [ ] Loading states and success/error notifications are displayed correctly.

- [x] Local state (profileData) reflects the backend response.

lockermanwxlf:
Fixes #344 

**What was changed?**

Submission and cancel buttons are disabled while posting

Success message shown on successful post, and local profile data is updated to server response.

Local profile data is initially queried from server.

Form submission now POSTS to server.

Acceptance Criteria:

- [x] Submitting the Profile Edit form sends a real POST request to /volunteer/profile.

- [x] The volunteer’s profile is created or updated correctly on the backend.

- [x] The read-only email field is never sent in the request.

- [x] Loading states and success/error notifications are displayed correctly.

- [x] Local state (profileData) reflects the backend response.

**Why was it changed?**

#344 

**How was it changed?**

Local form data is mapped to the request body the server expects. Mapped data is posted to endpoint.

### DIFF
--- a/client_app/src/api/volunteerApi.js
+++ b/client_app/src/api/volunteerApi.js
@@ -1,4 +1,4 @@
-import { fetchClient, patchRequest } from "./fetchClient";
+import { patchRequest } from "./fetchClient";
 import httpClient from "./httpClient";
 
 /**


lockermanwxlf:
Currently, server communicated errors are not shown to the user. For this specific use case, all validations from the server are also done (and displayed) on the client, but this seems to reflect a larger problem (unless I am using fetchClient wrong).

The issues I am having with this are:

* Currently, the backend POST endpoint returns a body like { errors: string[] }. fetchClient only supports a body like { message: string } or a string response.
* On error, fetchClient only raises an Error, from which we can only get a single string.

I think we can:

1. Wrap fetchClient's return value in a discriminated union which has support for typed errors (or make a new fetchClient API and migrate to it over time)
2. Add support for errorData.errors in fetchClient and just raise an error with the first error in it.
3. Not use fetchClient for this specific POST.

I think we should at least have some way of getting typed errors from our fetchClient implementation, as right now I'm just detecting a 404 with a try-catch, and for all I could be treating some other unrelated thrown error as a 404 down the line.

Or if we just change fetchClient to only perform request interception (and not wrap all non-200 codes in an exception), we can parse server errors more easily in the respective API classes/methods.

@Supraja050202

client_app/src/components/volunteer/Profile.js
```diff
@@ -55,15 +56,42 @@ const ProfileSettings = () => {
     skills: "",
   };
```
Supraja050202: Reorder hooks: declare all useState hooks before any useEffect calls for clarity and convention.

client_app/src/api/volunteerApi.js
```diff
@@ -1,13 +1,22 @@
-import { getRequest, patchRequest } from "./fetchClient";
+import { fetchClient, patchRequest } from "./fetchClient";
```
Supraja050202: fetchClient is imported but never used. Unused imports cause lint warnings.

Supraja050202 (approved):
Approved!

PR #372, opened 2025-12-01
Issue 366

Reva335:
Fixes #issue_366

**What was changed?**

*Here, describe what part of the application you changed (e.g. login page, database, etc.). If possible, mention what specific components were added, removed, or modified.*

**Why was it changed?**

*Here, describe the issue that you are fixing with this code, and why your code fixes it.*

**How was it changed?**

*Here, get into detail about what files you modified, and talk about the most important lines in regards to fixing the issue.*

### DIFF
--- a/client_app/src/formatting/FormatDateTime.js
+++ b/client_app/src/formatting/FormatDateTime.js
@@ -46,7 +46,7 @@ export const timeStringToMillis = (timeStr) => {
 
 // Convert timestamp to HH:mm format for time input
 export const timestampToTimeInput = (timestamp) => {
-  if (!timestamp) return "";
+  if (timestamp == null) return "";
   const date = new Date(timestamp);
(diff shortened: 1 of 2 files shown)


client_app/src/formatting/FormatDateTime.js
```diff
@@ -1,47 +1,70 @@
...
+}
 
 export const timeStringToMillis = (timeStr) => {
   const [hours, minutes] = timeStr.split(":").map(Number);
   return (hours * 60 + minutes) * 60 * 1000;
 };
 
+// Convert timestamp to HH:mm format for time input
+export const timestampToTimeInput = (timestamp) => {
+  if (!timestamp) return "";
```
Supraja050202: This is good defensive coding.
However, note that timestamp === 0 will incorrectly return "", even though midnight is valid time.
Consider updating to if (timestamp == null).

PR #401, opened 2026-04-13
Backend connection with frontend implemented for issue 394

### LINKED ISSUE #394: API Endpoint to Return Open Shelters List in Descending Date Order
name: Feature request
about: Suggest an idea for this project
title: API Endpoint to Return Open Shelters List in Descending Date Order
labels: enhancement, backend, api
assignees: ''

Is your feature request related to a problem?
The frontend requires a simple and efficient way to retrieve shelter availability data without relying on multiple requests or client-side filtering.

Describe the solution you'd like
Create a backend/API endpoint that returns a list of open shelters grouped by date in descending order (latest dates first), eliminating the need for date-based queries from the frontend.

Return shelter data grouped by date in descending order
Under each date, include all shelters open on that day
Include lightweight metadata (e.g., name and location)
Exclude shift-level schedule details
Provide all necessary data in a single response

Describe alternatives you've considered

API endpoint with day/date parameter filtering
Sending full weekly data and filtering in the frontend
Multiple API calls per shelter to determine availability

Additional context
Returning pre-grouped shelter data in descending date order simplifies frontend implementation and improves performance by reducing the need for additional filtering or multiple requests.

orhankoylu:
Fixes #394

What was changed?

Added a new backend API endpoint for open shelters and brought the related volunteer UI into this branch so the feature can be viewed end to end. The backend now exposes grouped open-shelter availability in a single response, and the volunteer dashboard includes the Open Shelters List page and navigation entry needed to display it.

Why was it changed?

Issue #394 asked for an API endpoint that returns open shelters grouped by date in descending order so the frontend does not need multiple requests or client-side grouping logic. This change fixes that by moving the grouping, deduplication, and date sorting into the backend response. I also brought over the volunteer UI from the previous branch so this branch can be tested in the app instead of only through a raw API call.

How was it changed?

Created server/use_cases/shelters/list_open_shelters_by_date_use_case.py to group future service shifts by date, dedupe shelters per date, sort dates descending, and serialize only lightweight shelter metadata. Updated server/application/rest/shelter.py to add GET /shelters/open, which collects future service shifts and returns the grouped response. Added a REST test in server/tests/rest/shelter/test_shelter.py to verify response shape, date ordering, per-date shelter deduplication, and future-only query behavior.

To make the branch testable in the UI, I also brought over the volunteer frontend files from the related work: client_app/src/components/volunteer/OpenSheltersCalendar.js, client_app/src/utils/openShelterCalendar.js, client_app/src/styles/volunteer/OpenSheltersCalendar.css, client_app/src/components/NavigationConfig.js, and client_app/src/App.js. Those files expose the Open Shelters List page in the volunteer dashboard so the backend work can be exercised from the app.

Screenshot 2026-04-13 at 11 28 18 AM

### DIFF
--- a/client_app/src/components/volunteer/OpenSheltersCalendar.js
+++ b/client_app/src/components/volunteer/OpenSheltersCalendar.js
@@ -1,12 +1,15 @@
-import { useEffect, useState } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 import { shelterAPI } from '../../api/shelter';
+import { serviceShiftAPI } from '../../api/serviceShift';
 import Loading from '../Loading';
 import { formatDate } from '../../formatting/FormatDateTime';
 import { ShelterInfo } from './ShelterInfo';
+import { getOpenSheltersGroupedByDate } from '../../utils/openShelterCalendar';
 import '../../styles/volunteer/OpenSheltersCalendar.css';
 
 function OpenSheltersCalendar() {
-  const [openShelterGroups, setOpenShelterGroups] = useState([]);
+  const [shelters, setShelters] = useState([]);
+  const [futureShifts, setFutureShifts] = useState([]);
   const [loading, setLoading] = useState(true);
   const [loadError, setLoadError] = useState('');
 
@@ -14,8 +17,13 @@ function OpenSheltersCalendar() {
     const fetchCalendarData = async () => {
       setLoadError('');
       try {
-        const groupedShelters = await shelterAPI.getOpenShelters();
-        setOpenShelterGroups(groupedShelters);
+        const [shelterData, shiftData] = await Promise.all([
+          shelterAPI.getShelters(),
+          serviceShiftAPI.getFutureShifts(),
+        ]);
+
+        setShelters(shelterData);
+        setFutureShifts(shiftData);
       } catch (error) {
         console.error('Error loading open shelters calendar:', error);
         setLoadError('We could not load the open shelters list right now. Please try again.');
@@ -26,6 +34,11 @@ function OpenSheltersCalendar() {
 
     fetchCalendarData();
   }, []);
+
+  const openShelterGroups = useMemo(
+    () => getOpenSheltersGroupedByDate(shelters, futureShifts),
+    [futureShifts, shelters]
+  );
 
   if (loading) {
     return <Loading />;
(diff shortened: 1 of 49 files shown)


server/application/rest/shelter.py
```diff
@@ -37,6 +43,25 @@ def get_shelters():
     )
```
Supraja050202, line 45: If `service_shifts_list_use_case()` or `shelter_list_use_case()` fails, this endpoint will return an unhandled 500 error without a structured response.This makes it difficult for the client to handle failures and provides no useful debugging context.
Wrap the logic in a try-except block and return a consistent error response.

client_app/src/components/volunteer/OpenSheltersCalendar.js
```diff
@@ -0,0 +1,83 @@
...
+import { formatDate } from '../../formatting/FormatDateTime';
+import { ShelterInfo } from './ShelterInfo';
+import { getOpenSheltersGroupedByDate } from '../../utils/openShelterCalendar';
+import '../../styles/volunteer/OpenSheltersCalendar.css';
+
+function OpenSheltersCalendar() {
+  const [shelters, setShelters] = useState([]);
+  const [futureShifts, setFutureShifts] = useState([]);
+  const [loading, setLoading] = useState(true);
+
```
Supraja050202, line 14: Missing error state and UI
When the API call fails:
- Error is only logged to console
- No error state is set
- UI renders empty/blank content
This results in poor user experience with no indication of failure.
Add an error state and render fallback UI.

client_app/src/components/volunteer/OpenSheltersCalendar.js
```diff
@@ -0,0 +1,83 @@
+import { useEffect, useMemo, useState } from 'react';
+import { shelterAPI } from '../../api/shelter';
+import { serviceShiftAPI } from '../../api/serviceShift';
+import Loading from '../Loading';
+import { formatDate } from '../../formatting/FormatDateTime';
+import { ShelterInfo } from './ShelterInfo';
+import { getOpenSheltersGroupedByDate } from '../../utils/openShelterCalendar';
+import '../../styles/volunteer/OpenSheltersCalendar.css';
+
+function OpenSheltersCalendar() {
```
Supraja050202, line 8: Redundant API calls
This component is making two API calls:
- shelterAPI.getShelters()
- serviceShiftAPI.getFutureShifts()
Then grouping data client-side using getOpenSheltersGroupedByDate().
However, the new backend endpoint `/shelters/open` introduced in this PR already performs this grouping.
This causes
- Unnecessary data transfer
- Duplicate processing logic
- Increased frontend complexity
- Performance overhead
Call the backend endpoint directly instead of fetching raw data.

Supraja050202 (changes requested):
Great work —this PR delivers clear functionality and is well-structured overall.
However, there are a few high-priority issues to address, including missing backend error handling and redundant API calls in the frontend.
The frontend should rely on the new /shelters/open endpoint to avoid duplicate processing and improve performance.
Additionally, error states are not handled in the UI, which can result in a blank screen on failure.
Addressing these will significantly improve reliability, performance, and user experience.

Supraja050202 (approved):
Approved!

